PR #1, opened 2021-04-19
Use smart pointers

kpu:


mainwindow.h
```diff
@@ -36,8 +36,8 @@ private slots:
     void on_localModels_activated(int index);
 
 private:
-    Ui::MainWindow *ui;
-    MarianInterface * translator_; // Initialised to nullptr. Can't use smart ptrs due to destructor needing to run
+    std::unique_ptr<Ui::MainWindow> ui_;
```
XapaJIaMnu, line 39: Could we use `QPointer` for QObjects?

mainwindow.h
```diff
@@ -36,8 +36,8 @@ private slots:
...
+    std::unique_ptr<MarianInterface> translator_;
```
XapaJIaMnu, line 40: I disabled this, because we can't use `.reset()`, as it first creates a new object, later destroys old one and one could easily crash the application if they don't know that you can't use this method.

mainwindow.h
kpu: QPointer doesn't delete, it just lets you know the object was deleted.  

https://stackoverflow.com/questions/40346393/should-i-use-qscopedpointer-or-stdunique-ptr

XapaJIaMnu: Ok, then I should refactor the whole code to not use the (useless) QPointer, but to instead use unique_ptr

PR #3, opened 2021-04-22
Produce cleaner dmg volume name by running macdeployqt with local path

jelmervdl:
See https://doc.qt.io/qt-5/macos-deployment.html#volume-name: "The volume name of a disk image (the text displayed in the window title of an opened .dmg file) created with -dmg is based on the path to the application when macdeployqt is run."


XapaJIaMnu:
Thank you very much!

PR #10, opened 2021-05-06
Reduce the number of interface elements

jelmervdl:


MarianInterface.cpp
```diff
@@ -25,22 +25,34 @@ marian::Ptr<marian::Options> MakeOptions(QString path_to_model_dir) {
...
         std::future<marian::bergamot::Response> responseFuture = service_->translate(std::move(input));
         responseFuture.wait();
         marian::bergamot::Response response = responseFuture.get();
+
+        // There is no guarantee that we get/process responses in the same order
+        // as we sent sentences to be translated. So let's make sure we haven't
+        // been overtaken by a further progressed sentence before emitting the
+        // translation.
+        // TODO: race condition on finished_?
+        if (serial < finished_)
```
XapaJIaMnu, line 48: Not necessary? We block until the translation is done.

ModelManager.cpp
```diff
@@ -5,16 +5,28 @@
...
+#include <QJsonObject>
+#include <QJsonArray>
+#include <QNetworkAccessManager>
+#include <QNetworkReply>
 #include <iostream>
 // libarchive
 #include <archive.h>
 #include <archive_entry.h>
 
+const int ModelManager::kColumnName = 0;
```
XapaJIaMnu, line 18: Some comments here?

mainwindow.h
```diff
@@ -49,8 +56,14 @@ private slots:
 
     // Model and config manager
     ModelManager models_;
+    ModelListItemDelegate localModelDelegate_;
 
     // Network code:
     Network network_;
+
+    QTimer inactivityTimer_;
```
XapaJIaMnu, line 64: Is this timeout for qt < 5.15?

XapaJIaMnu (commented):
A few questions.

XapaJIaMnu:
We should fix the constexpr thingies

PR #12, opened 2021-05-07
Reduce number of interface elements: clean-up

jelmervdl:
- Clean up `static const` code in favour of enums + constexpr
- Add menu item to disable *translate as you type* in favour of a plain old "translate!" button. On macOS cmd + enter also works to trigger translation. I assume on Linux/Windows this translates to super + enter.
- Make distinction between *no model list downloaded* and *no models available that you don't already have locally installed*.
- Clearing out the input text box now also properly clears out the translated output box
- Removed "Translating…" text that shows up in the translation output box for now as it was constantly overridden with partial translations. This should be replaced by a separate loading indicator. I'll open a separate issue to discuss this because there's plenty of options but none stand out to me as the obvious one. Update: #13


XapaJIaMnu:
@jelmervdl on linux/windows it is ctrl + enter (the execute translation shortcut)

Thank you for the great work here!

Issue #11, opened 2021-05-06
Error: Unhandled exception of type 'St12length_error': basic_string::_M_create


kpu:
```
[2021-05-06 20:04:54] Error: Unhandled exception of type 'St12length_error': basic_string::_M_create
[2021-05-06 20:04:54] Error: Aborted from void unhandledException() in /home/kpu/translateLocally/3rd_party/bergamot-translator/3rd_party/marian-dev/src/common/logging.cpp:113

[CALL STACK]
[0x5588f304fd0f]                                                       + 0x64d0f
[0x7f4e083ae7c6]                                                       + 0xa67c6
[0x7f4e083ae831]                                                       + 0xa6831
[0x7f4e083aeac3]                                                       + 0xa6ac3
[0x7f4e083a54e1]    std::  __throw_length_error  (char const*)         + 0x3d
[0x7f4e08443c8b]                                                       + 0x13bc8b
[0x5588f5640a71]                                                       + 0x2655a71
[0x5588f56410ee]                                                       + 0x26560ee
[0x5588f5641855]                                                       + 0x2656855
[0x5588f310f5b0]                                                       + 0x1245b0
[0x5588f30fadea]                                                       + 0x10fdea
[0x5588f30fcf22]                                                       + 0x111f22
[0x5588f30fcf96]                                                       + 0x111f96
[0x5588f30ccf01]                                                       + 0xe1f01
[0x7f4e083d9ef0]                                                       + 0xd1ef0
[0x7f4e09275f5e]                                                       + 0x7f5e
[0x7f4e080f41ef]    clone                                              + 0x3f

Aborted (core dumped)
```

Steps to reproduce:
1. Commit https://github.com/jelmervdl/translateLocally/commit/3f81d83e4d7839eac99f2b5e640dbb5503f9afd0
2. Patch the annoying pcre2 bug to get anything to compile
```
diff --git a/CMakeLists.txt b/CMakeLists.txt
index 0319750..1b17bc0 100644
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -72,8 +72,8 @@ if(WIN32) # Windows static extension is .lib
         set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -fPIC -O3")
     endif(MSVC)
 else(WIN32)
-    set(CMAKE_FIND_LIBRARY_SUFFIXES .a)
[...]

PR #17, opened 2021-05-10
Move interaction with marian into single thread

jelmervdl:
Work-around for #11 which seems to be caused by multiple threads accessing the batcher concurrently. This also moves most of Marian's start up of the main thread.

There is some room for improvement before merging: the destructor waits for marian to finish. When switching models at start-up of translateLocally, this causes the GUI to hang for a few seconds. Ideally we could just detach the spinning down worker thread (or MarianInterface object) and have it clean up itself.

Re @XapaJIaMnu in #15: This change contains a "queue" of at most 1 pending text, which gets replaced with a newer input text if marian is still chewing on the previous input text. This way there's never more than three copies of input text floating around: one in the input text field, one in the input queue of MarianInterface, one in the worker (shared with Marian). (And one in the MainWindow used to determine whether the input changed enough to warrant sending it for translation again.)

### DIFF
--- a/MarianInterface.cpp
+++ b/MarianInterface.cpp
@@ -4,7 +4,9 @@
 #include "3rd_party/bergamot-translator/src/translator/parser.h"
 #include "3rd_party/bergamot-translator/src/translator/response.h"
 #include "3rd_party/bergamot-translator/3rd_party/marian-dev/src/3rd_party/spdlog/spdlog.h"
+#include <memory>
 #include <thread>
+
 
 namespace  {
 marian::Ptr<marian::Options> MakeOptions(QString path_to_model_dir, translateLocally::marianSettings& settings) {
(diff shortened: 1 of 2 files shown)


MarianInterface.cpp
```diff
@@ -28,58 +30,57 @@ marian::Ptr<marian::Options> MakeOptions(QString path_to_model_dir, translateLoc
...
-        // If translating empty string, just pass through immediately but do
-        // update the finished_ variable to cancel out all previous translations.
-        if (input.empty()) {
-            translation = "";
-        } else {
-            std::future<marian::bergamot::Response> responseFuture = service_->translate(std::move(input));
+    worker_ = std::thread([&](marian::Ptr<marian::Options> options) {
+        marian::bergamot::Service service(options);
+
+        while (true) {
```
XapaJIaMnu, line 40: This looks like busy waiting?

XapaJIaMnu: So, it's a semaphore acquiring. Is there a chance for this to busy-wait? When we have a large translation pending and the user tries to submit a new one?

XapaJIaMnu: Ok, nvm, github is showing the diff in a weird way.

So basically, you bypass the service queuing mechanism by creating your own worker and queue. Which is bad because we're reimplementing functionality from a broken upstream service we're using. But I understand

MarianInterface.cpp
```diff
@@ -28,58 +30,57 @@ marian::Ptr<marian::Options> MakeOptions(QString path_to_model_dir, translateLoc
...
 MarianInterface::~MarianInterface() {
-    // We need to manually destroy the loggers, as marian doesn't do that.
-    spdlog::drop("general");
-    spdlog::drop("valid");
+    // Poision the worker with a nullptr
+    std::unique_ptr<std::string> old(pendingInput_.fetchAndStoreAcquire(nullptr));
+    if (!old)
+        pendingInputCount_.release();
+    
+    // Wait for worker to join as it depends on resources we still own.
```
XapaJIaMnu, line 79: This is not an issue in the current implementation as we do a thread::detach, but potentially could lead to crashes. I think we disable (or at least I used to) disable changing the model while translation is still ongoing...

XapaJIaMnu:
So, jerin is rewriting bergamot-translator to use a callback, and not a future. That could potentially help, but I think that we want to drop bergamot-translator and write our own queue management on top of vanilla marian.

MarianInterface.cpp
jelmervdl: I'm tempted to add a "detachAndFree()" method or something to MarianInterface. That then basically starts a separate thread that does nothing but `delete this`. When changing model, we could do something like `translator_.release()->detachAndFree()` and just let the MarianInterface object take care of cleaning up itself in a non-main thread-blocking way.

XapaJIaMnu:
Regardless, this is a 100% improvement from the previous behaviour, @kpu do you think we should merge this while we implement a proper solution?

MarianInterface.cpp
XapaJIaMnu: That means the worker thread must somehow know it's not "relevant" so that it doesn't emit any signals that mess up the GUI.

XapaJIaMnu: Yepp, this is the spdlog issue that i have in the destructor.

PR #22, opened 2021-05-12
Move all marian interactions into worker

jelmervdl:
I've moved the (re)creation of `marian::bergamot::Service` into the worker itself, and added a `MarianInterface::setModel()` method.
- This removes the need to recreate the `MarianInterface` when switching models. (Yay no more `disconnect()` & `connect()`!)
- It organises the destruction of the old and construction of the new `marian::bergamot::Service` and `marian::Config` instances such there are never two at the same time. Because that causes trouble around the logging interface.
- Additionally, this moves the loading & destruction off the main thread keeping the interface "snappy".
- And I now throw & catch errors so loading a broken model no longer crashes the application.
- As a bonus, this change also fixes a possible race condition in my previous implementation.
- Finally I added some notes on why this design is as it is.
- Update: removed disabling translate button while translating as there is no technical reason to do so.

### DIFF
--- a/mainwindow.cpp
+++ b/mainwindow.cpp
@@ -25,6 +25,7 @@ MainWindow::MainWindow(QWidget *parent)
     , localModelDelegate_(this)
     , translatorSettingsDialog_(this, models_.getSettings())
     , network_(this)
+    , translator_(new MarianInterface(this))
 {
     ui_->setupUi(this);
     ui_->statusbar->addPermanentWidget(ui_->pendingIndicator);
@@ -41,9 +42,6 @@ MainWindow::MainWindow(QWidget *parent)
 
     updateLocalModels();
 
-    if (!models_.installedModels().empty())
-        resetTranslator(models_.installedModels().first().path);
-
     inactivityTimer_.setInterval(300);
     inactivityTimer_.setSingleShot(true);
     
@@ -54,6 +52,15 @@ MainWindow::MainWindow(QWidget *parent)
     connect(&models_, &QAbstractTableModel::rowsRemoved, this, &MainWindow::updateLocalModels);
     connect(&network_, &Network::error, this, &MainWindow::popupError); // All errors from the network class will be propagated to the GUI
     connect(&translatorSettingsDialog_, &TranslatorSettingsDialog::settingsChanged, this, &MainWindow::updateModelSettings);
+
+    // Set up the connection to the translator
+    connect(translator_.get(), &MarianInterface::pendingChanged, ui_->pendingIndicator, &QProgressBar::setVisible);
+    connect(translator_.get(), &MarianInterface::translationReady, this, [&](QString translation) {
+        ui_->outputBox->setText(translation);
+        ui_->localModels->setEnabled(true); // Re-enable model changing
+        ui_->translateAction->setEnabled(true); // Re-enable button after translation is done
+        ui_->translateButton->setEnabled(true);
+    });
 
     // Queue translation when user has stopped typing for a bit
     connect(&inactivityTimer_, &QTimer::timeout, this, [&] {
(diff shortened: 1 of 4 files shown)


mainwindow.cpp
```diff
@@ -209,31 +220,7 @@ void MainWindow::translate(QString const &text) {
...
-    connect(translator_.get(), &MarianInterface::pendingChanged, ui_->pendingIndicator, &QProgressBar::setVisible);
-    connect(translator_.get(), &MarianInterface::translationReady, this, [&](QString translation) {
-        ui_->outputBox->setText(translation);
-        ui_->localModels->setEnabled(true); // Re-enable model changing
-        ui_->translateAction->setEnabled(true); // Re-enable button after translation is done
-        ui_->translateButton->setEnabled(true);
-    });
-
+    translator_->setModel(dirname + "/", models_.getSettings());
     translate();
```
XapaJIaMnu, line 217: We always translate after changing the model?

jelmervdl: Yes, to show that the new model has been loaded and have the translation show up in the new language. If the current input is empty string it will just sail right through without much overhead.

mainwindow.cpp
```diff
@@ -56,12 +56,7 @@ MainWindow::MainWindow(QWidget *parent)
     // Set up the connection to the translator
     connect(translator_.get(), &MarianInterface::pendingChanged, ui_->pendingIndicator, &QProgressBar::setVisible);
     connect(translator_.get(), &MarianInterface::error, this, &MainWindow::popupError);
-    connect(translator_.get(), &MarianInterface::translationReady, this, [&](QString translation) {
-        ui_->outputBox->setText(translation);
-        ui_->localModels->setEnabled(true); // Re-enable model changing
-        ui_->translateAction->setEnabled(true); // Re-enable button after translation is done
-        ui_->translateButton->setEnabled(true);
-    });
+    connect(translator_.get(), &MarianInterface::translationReady, ui_->outputBox, &QTextEdit::setText);
```
XapaJIaMnu, line 59: What happens if you issue multiple translation commands now? Eg you're translating war and peace and it's takign too long and you as the impatient user issue an extra translation request? Would that queue?

jelmervdl: Yes, but only the last request will be queued.

You can have one translation in progress (that then only lives in the worker thread) and one pending translation (the `MarianInterface::pendingInput_` pointer). While marian is still busy translating, queueing more translations will only replace the `pendingInput_` one.

XapaJIaMnu: I don't like this because it could accidentally queue an extra long translation that I wouldn't want and there's no way to cancel it. I prefer a new translation to only start once the old one is complete, at least until we can cancel translations?

jelmervdl: > I don't like this because it could accidentally queue an extra long translation that I wouldn't want and there's no way to cancel it.

You mean something like submitting a form multiple times? That may be better solved by disabling the button as long as the input text remains unchanged. (This would double as an indicator that the translation that pops up after N seconds is actually of the text that you originally typed.)

I imagined the scenario where you pasted in a long email, push "Translate", realise that you forgot to sign it off or something and add that text while it's translating. If the translate button is disabled while translating, you then have to actively wait for it to be enabled again, don't forget to click it, and wait again. If you can always click & queue, you just have to wait once (although for about the same amount of time.)

XapaJIaMnu: Could we change it back to make sure we  only queue the latest translation.

jelmervdl: I'm not sure I understand, that's what it does now: when you press "Translate" you replace the queued translation input (or a nullptr if nothing is queued).

Do you mean disable the Translate button while translating is in progress, and only enable it when marian is available again?

jelmervdl: I've reverted this UI behaviour.

Issue #23, opened 2021-05-12
Persist settings


XapaJIaMnu:
We should persist the settings on relaunch. Load the same model that we had (if it exists), same number of cores, same memory, same input/output text (controllable by the settings dialog)

XapaJIaMnu:
Yeah, some of that was quite hacky... The point I was going for is that if we have everything in QSettings, then saving and loading is going to be very easy, but I didn't develop it much.

jelmervdl:
I thought it would be easy this way, but discovered that while QtQuick and Qt6 have bidirectional property binding, in Qt5/C++ you have to do it manually 😞 So you still end up writing a bunch of `connect()`s just to keep it all in sync.

XapaJIaMnu:
Shame....

XapaJIaMnu:
We do have persistent settings now, but...
If you wipe the models, and try to start translating, you would be greeted with a confusing error message that "you need to download models before you start translating". This is due to the settings attempting to restore the typed in text.

Furthermore, we should allow the user to wipe all typed in text on exit, to prevent some embarassing start up screens.

jelmervdl:
I'll fix the first error by not triggering translation on start-up (basically block the textChanged signal caused by restoring input text.)

This also saves users from triggering a large translate job on start-up. I can imagine a scenario where you trigger translation, then realise this is taking too long, try to close the program. Then, after the program finally closed (because you didn't force-close it or anything) and you open it again, it restores that long text and begins translating! Ugh.

Re "allow the user to wipe all typed in text on exit" should this
1. just be a checkbox in the settings pane that lets you turn on/off remembering the last input text?
2. Or a separate menu option to quit the program and forget the input text.
3. Could be a combination of both. Have a setting for default behaviour, and still have a menu item to trigger the other behaviour. This is pretty common in document-based macOS apps (e.g. Pages) that remember your open windows when you quit them. There's a global preference in macOS that tells apps they should remember their last open application state. With alt + cmd + q you override this behaviour when closing an app.

One more consideration: should this clear option overwrite the last saved text with nothing, or leave it untouched? I think clearing it makes more sense.

kpu:
For a privacy-focused product, I think it's really off-brand to be storing text translated on disk.  In my opinion, empty should be the default.

jelmervdl:
Hmm, that's true.

[...]

Issue #18, opened 2021-05-10
When a new model is download, we do require an addition selection to switch to that language pair


XapaJIaMnu:
@jelmervdl this would need to be fixed. I'll do it as I refactor the model selection.

jelmervdl:
In `MainWindow::handleDownload()`, adding a call to `MarianInterface::resetModel` after `models_.writeModel(...)` should do the trick. Maybe `writeModel` should return whether or not it was successful, so you don't call resetModel on a model that never got extracted correctly.

jelmervdl:
I quickly tried to see whether it was really as simple as I imagined. it wasn't, I forgot that you would also need to select the newly downloaded model in the model dropdown. And since you're refactoring that part anyway, I wasn't planning on merging this. Still, in case it may be of any use to you: https://github.com/jelmervdl/translateLocally/tree/select-downloaded-model

XapaJIaMnu:
I will fix it when I finish the refactoring, don't worry about it for now.

PR #25, opened 2021-05-20
Persistent settings

jelmervdl:
Implementation for #23 (and accidentally also #18…)

This adds a class `Settings` that describes the state of the application.
1. Each setting is described by a `Setting` class which signals changes. Other parts of the application can listen to these changes to update UI and translation model.  
Technically this is split up in `Setting(extends QObject)` and `SettingImpl<T>(extends Setting)`. This because you can't use Q_OBJECT (and thus signals) with template classes, but I did want to use templates to describe the data type of settings.
2. Most settings are saved & applied when they change (synced through signals). Exception is the inputText (and cursor position) as these change quite often, but have little impact across the application. Therefore these are only loaded & saved on application startup & shutdown.
3. It stores both input text and cursor position (to give the impression you can just pick up where you left off last time). ~~I do not trigger translation on restoring the input text when `translateImmediately` is enabled to prevent accidental lengthy translation processes. Once cancelling a translation is implemented, this should probably change.~~ `translateImmediately? translate()` will trigger on load as well because restoring the text emits the `inputBox_textChanged` signal.
4. I changed `ModelManager::writeModel` to return the model so it can be automatically selected when it finishes downloading & extracting. 
5. Changed the dropdown option generation in the settings dialog a bit to use the userData field of the options. Makes the code a bit more copy/pastable (or easier to move around).

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -106,6 +106,8 @@ set(PROJECT_SOURCES
         ModelManager.h
         Network.cpp
         Network.h
+        Settings.cpp
+        Settings.h
         types.h
         ${TS_FILES}
 )
--- a/MarianInterface.cpp
+++ b/MarianInterface.cpp
@@ -11,7 +11,7 @@
 
 namespace  {
 marian::Ptr<marian::Options> MakeOptions(const std::string &path_to_model_dir, translateLocally::marianSettings& settings) {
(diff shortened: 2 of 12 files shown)


XapaJIaMnu:
Ugh, merging time. I started  working on Unifying all the models and the JSON parsers so now need to merge...

Issue #4, opened 2021-04-22
Equal size vsplit windows for source and target


jerinphilip:
Feature request:

![image](https://user-images.githubusercontent.com/727292/115758446-08ebea80-a397-11eb-9ab9-cfab95c5dd13.png)

Can I have an equal size vsplit windows for source and target text?

Wishful thinking - synchronized scrolling.

XapaJIaMnu:
You can't have equal size. What you want to have is adjustable size. Which is on the TODO

jerinphilip:
> You can't have equal size.

Confused here, Why? I want only one control that adjusts size, and adjusts size of both by the same value scale-x, scale-y on x, y. Both should be scrollable in overflow of text (and ideally sync-scrollable). 

vsplit to check source and translation side by side.

XapaJIaMnu:
ok, let me be more explicit. This would be so far down the priorities that I don't want to consider it a potential feature for now.

Why? Too many options. You want locked scrolling? Sure, but maybe some users don't want that. You want the window sizes synchronised? Most people don't really want to waste that much screen real estate on the input as they don't really care about it, especially if it's a language they can't even read (eg Chinese). Adding all these options in an intuitive and unobstructive way  so that it doesn't clutter the GUI would take quite a bit of time.

Potentially in the future, you could use the attention mechanism to highlight words from the two interfaces, auto scroll and so on, but for now I think if I make the size of the fields adjustable, it would solve 99% of the use cases.

Maybe side-by-side arangment like google translate is a good idea, even if I personally don't like it, as I don't often care about the input.

jelmervdl:
Currently we have a (resizable) vertical split. Adding a toggle to the menu to change this to a horizontal split would be quite easy I presume. (but then we should also save the window size in the settings so that people who toggle it to horizontal don't have to resize the window every time) Is that something we would like?

Synced scrolling: implementing it in a naive way is not difficult, but could be pretty annoying for users when it doesn't work. Should be a toggle in the menu. I'm tempted to add this only when the text field is more aware of sentences, and then have both fields optionally scroll both source and target sentences to be visible instead of relative scroll position.

jerinphilip:
> Synced scrolling: implementing it in a naive way is not difficult, but could be pretty annoying for users when it doesn't work. Should be a toggle in the menu.

[...]

PR #28, opened 2021-05-24
Horizontal split option

jelmervdl:
Implementation for the horizontal split suggested in #4.

# Description
This change adds
1. A "View" menu option that lets you toggle between horizontal and vertical splitting the input/output boxes
2. Toggle between them with ctrl+1/ctrl+2
2. Saving & restoring this split preference across restarts
3. Saving & restoring window geometry (size, position) across restarts
4. Add `bind()` template which is `connect()` plus the initial update on startup. To get a bit closer to a proper property binding programming experience.

Screenshot 2021-05-24 at 11 03 35
(Note: The "Enter Fullscreen" menu item is something added by macOS to all View menus in all applications.)


XapaJIaMnu:
Thank you for the hard work, I wasn't aware that this could be made optional with so little code!

PR #29, opened 2021-05-24
Add a speed indicator in the status bar

XapaJIaMnu:
I'm not sure how to avoid the string copy when counting words without a major refactoring? Any ideas, could I do it better?
Screenshot_20210524_170742

### DIFF
--- a/src/mainwindow.cpp
+++ b/src/mainwindow.cpp
@@ -38,7 +38,9 @@ MainWindow::MainWindow(QWidget *parent)
     this->setWindowIcon(icon);
 
     // Create the status bar
+    ui_->statusbar->addWidget(ui_->speedDisplay);
     ui_->statusbar->addPermanentWidget(ui_->pendingIndicator);
+    ui_->speedDisplay->hide();
     ui_->pendingIndicator->hide();
 
     // Hide download progress bar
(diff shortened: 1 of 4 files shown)


src/mainwindow.cpp
```diff
@@ -66,10 +68,16 @@ MainWindow::MainWindow(QWidget *parent)
     // Set up the connection to the translator
     connect(translator_.get(), &MarianInterface::pendingChanged, ui_->pendingIndicator, &QProgressBar::setVisible);
     connect(translator_.get(), &MarianInterface::error, this, &MainWindow::popupError);
-    connect(translator_.get(), &MarianInterface::translationReady, this, [&](QString translation) {
+    connect(translator_.get(), &MarianInterface::translationReady, this, [&](QString translation, int speed) {
         ui_->outputBox->setText(translation);
         ui_->translateAction->setEnabled(true); // Re-enable button after translation is done
         ui_->translateButton->setEnabled(true);
+        if (speed > 0) { // Display the translation speed only if it's > 0. This prevents the user seeing weird number if pressed translate with empty input
+            ui_->speedDisplay->setText(QString("Translation speed: ") + QString::fromStdString(std::to_string(speed)) + QString(" words per second."));
```
jelmervdl, line 76: With QString you can do things like `QString("Translation speed: %1 words per second").arg(speed)`

jelmervdl:
I agree, I don't see an easy way to avoid the copy. bergamot-translate's Service::translate takes ownership, and it doesn't make much sense to change that given its interface with the std::future.

I initially thought "hey you could just do the counting in the worker thread, that has access to input, and would be just waiting otherwise anyway, right?!" but yeah it transfers that ownership to translate() so even if you would do counting in the worker thread, you still don't have access to the string.

Is counting much slower than a copy? If not, you could do the counting before giving the input string to it to translate(). No need for the std::async thread pool in the background. It already happens in the worker thread so it would not block the UI.

XapaJIaMnu:
I am not sure if the counting is slower than the copy. It's a linear sweep. In theory the copy should be faster as you know the memory size in advance...

XapaJIaMnu:
In the future maybe we can outsource this to the service.

jerinphilip:
Give `Service` an update-callback which will be called and updates UI when n sentences in a request are updated. https://github.com/browsermt/bergamot-translator/blob/22a1b9113ec983aac12c6ddb48a78d7ef8ebfe57/src/translator/request.cpp#L37-L47

We are switching to callbacks.

XapaJIaMnu:
I'm not sure how this fixes the string ownership issue..?

PR #33, opened 2021-05-27
Add LibArchive dependency as an ExternalProject

jelmervdl:
This change adds libarchive as an external cmake project to work around having to install it as a separate dependency when compiling.

Note: this change does not remove the installation of libarchive(-dev) from the Github workflows.


XapaJIaMnu:
Fixed by https://github.com/XapaJIaMnu/translateLocally/commit/b7e32804d8c4aa95c46db7c4ed2998a8b799b429

PR #43, opened 2021-06-04
Polish fetch remote list

### LINKED ISSUE #42: One minor bug: the language pair lists appears three times (Ubuntu 20.04.01).
@mlforcada says "One minor bug: the language pair lists appears three times (Ubuntu 20.04.01)." https://twitter.com/MikelForcada/status/1400537992343834628

jelmervdl:
- Make fetchRemoteModels reentrant (I think that's the term? Meaning you can call it multiple times it will still do only one request at a time)
- Replace the "Download models…" option with a "Downloading model list…" message while we're downloading said list.

This should fix #42.


kpu:
idempotent?

Issue #20, opened 2021-05-11
Change the way VERSION is parsed


XapaJIaMnu:
For some reason, on mac, if you have boost installed via homebrew, compilation will fail unless the file `VERSION` is removed. Investigate and fix.

XapaJIaMnu:
Some improvements have happened, at least for now there is a clean compilation on mac.

jelmervdl:
I no longer have any issues building on mac. As far as I'm concerned this issue is fixed.

XapaJIaMnu:
I want to change it a bit so that it gets baked into the executable

XapaJIaMnu:
@jelmervdl should we close this now?

PR #46, opened 2021-06-08
Use version from VERSION for `app -v` and macOS bundle version

jelmervdl:
Use version from VERSION for `app -v` and macOS bundle version. I've bumped the version in VERSION to match the one that was hard-coded in the application. This should solve #20.


.gitignore
```diff
@@ -75,3 +75,4 @@ build*/
 
 # Qt creator:
 CMakeLists.txt.user
+/src/version.h
```
XapaJIaMnu, line 78: This should go into `CMAKE_CURRENT_BINARY_DIR` not the source dir

CMakeLists.txt
```diff
@@ -113,6 +113,9 @@ set(PROJECT_SOURCES
         ${TS_FILES}
 )
 
+# Make PROJECT_VERSION_* available to our code as TRANSLATELOCALLY_VERSION_*
+configure_file(${CMAKE_CURRENT_SOURCE_DIR}/src/version.h.in
```
XapaJIaMnu, line 117: The generated file should go into the binary dir, not the source dir.

XapaJIaMnu (changes requested):
THank you for fixing this! I think the generated files at compile time should go into the binary dir.

jelmervdl:
I agree. I copied what Marian did, but the cmake documentation also seems to recommend using `${CMAKE_CURRENT_BINARY_DIR}`.

PR #47, opened 2021-06-08
Add User-Agent header to http requests

jelmervdl:
This adds a User-Agent header on all requests. The only use case for this change should be analytics:
- differentiate between browser, translateLocally, and any other programs that might be downloading models. (You can already do this right now by looking at the lack of a User-Agent header 😛)
- to get an idea which versions of translateLocally are in use, at least with the audience that downloads/updates models.


XapaJIaMnu (approved):
Looks good

Issue #45, opened 2021-06-04
Local model management


jelmervdl:
From #34, @XapaJIaMnu mentioned we should have some way to manage local models from the app (without going to a hidden folder)

I suggest adding a table view to the settings pane, as a second tab, where you can see the name of the model, the version number, the total file size, and a button to delete it.

Additionally, how do we expect users to download models if not though the app? Possible options:
1.  Add an "Import model…" menu item that opens a file browser dialog.
2. Add an "Import model…" button below the table in the settings pane that does the same.
3. Drag & drop .tar.gz files onto this table
4. Drag & drop .tar.gz files onto the app window or dock icon (in case of macOS, not sure how other desktop environments normally deal with this)
5. Make up a new file extension, rename all online models with it, and register the app as a viewer for said file extension so you can basically double click to install. Added bonus is that browsers won't automatically try to extract the file after downloading because it's a recognised archive format. (I'm looking at you, Safari…)
6. Just load any model that's in the same directory as the application; imagine having a USB stick with the .exe and a bunch of model directories. Trouble right now is that we need to have extracted archives for models to be loaded.

Will we host models for download through the browser somewhere? For the purpose of having a fully offline installation, this might be good. I assume some page somewhere on translatelocally.com would be the most logical place. Technically, there is also not really a reason to combine a bunch of models into a single .tar.gz and host it as an "English -> Anything starter kit" or something.

XapaJIaMnu:
Hey,

That sounds reasonable, with regards to the model management tab. As for the rest.

1. Are expecting folders or archives? Atm I scan a folder and if a model_info.json appears, it's valid. I scan the current folder, as well as the settings directory.
2. Same as 1
3. Sounds nice, but should we be able to handle folders
4. Same as 3
5. I don't think we should be having a new extension.
6. Already works, as long as the models are extracted. The folder name doesn't matter, as long as it's in the same directory as the .exe/compiled file.

XapaJIaMnu:
Fixed via #48

PR #48, opened 2021-06-08
Manage local models through application

jelmervdl:
Solving #45.

Idea is to:
- [x] Add a table to the settings dialog to see & delete local models
  - [x] only the ones in .config/translateLocally can be deleted. Do not attempt to delete models found in `pwd`.
  - [x] Make delete option prominent so the intention of that table becomes immediately clear  
_I'm thinking adding a column with a delete button per row, but might end up with a button at the bottom of the table that is only enabled if deletable modes are selected. Like how the context menu works. Because of space._
- [x] Add an import button to load (copy) local models into the local "library" (the .config/translateLocally folder)
  - [x] validate before importing (check whether there is a model_json file?)  
_partially done, model is not imported but extracted files stay_

This won't affect:
- Reading local models from the start-up directory. That should keep working. Those models should not be imported into the local library.

### DIFF
--- a/src/TranslatorSettingsDialog.ui
+++ b/src/TranslatorSettingsDialog.ui
@@ -121,12 +121,20 @@
        </property>
        <item>
         <widget class="QTableView" name="localModelTable">
+         <property name="contextMenuPolicy">
+          <enum>Qt::ActionsContextMenu</enum>
+         </property>
          <property name="selectionBehavior">
           <enum>QAbstractItemView::SelectRows</enum>
          </property>
          <property name="showGrid">
           <bool>false</bool>
          </property>
+         <attribute name="verticalHeaderVisible">
+          <bool>false</bool>
+         </attribute>
+         <addaction name="actionRevealModel"/>
+         <addaction name="actionDeleteModel"/>
         </widget>
        </item>
        <item>
@@ -168,6 +176,22 @@
     </widget>
    </item>
   </layout>
+  <action name="actionRevealModel">
+   <property name="text">
+    <string>Reveal folder</string>
+   </property>
+  </action>
+  <action name="actionDeleteModel">
+   <property name="text">
+    <string>Delete</string>
+   </property>
+   <property name="shortcut">
+    <string>Del</string>
+   </property>
+   <property name="shortcutContext">
+    <enum>Qt::WidgetShortcut</enum>
+   </property>
+  </action>
  </widget>
  <resources/>
  <connections>
@@ -178,8 +202,8 @@
    <slot>accept()</slot>
    <hints>
     <hint type="sourcelabel">
-     <x>20</x>
-     <y>20</y>
+     <x>32</x>
+     <y>354</y>
     </hint>
     <hint type="destinationlabel">
      <x>20</x>
@@ -194,8 +218,8 @@
    <slot>reject()</slot>
    <hints>
     <hint type="sourcelabel">
-     <x>20</x>
-     <y>20</y>
+     <x>32</x>
+     <y>354</y>
     </hint>
     <hint type="destinationlabel">
      <x>20</x>
--- a/src/TranslatorSettingsDialog.ui
+++ b/src/TranslatorSettingsDialog.ui
@@ -23,7 +23,7 @@
       <attribute name="title">
        <string>Behaviour</string>
       </attribute>
-      <layout class="QVBoxLayout" name="verticalLayout">
+      <layout class="QVBoxLayout" name="verticalLayout_1">
        <property name="leftMargin">
         <number>8</number>
        </property>
@@ -106,7 +106,7 @@
       <attribute name="title">
        <string>Languages</string>
       </attribute>
-      <layout class="QVBoxLayout" name="verticalLayout">
+      <layout class="QVBoxLayout" name="verticalLayout_3">
        <property name="leftMargin">
         <number>8</number>
        </property>
--- a/src/TranslatorSettingsDialog.ui
+++ b/src/TranslatorSettingsDialog.ui
@@ -6,63 +6,155 @@
    <rect>
     <x>0</x>
     <y>0</y>
-    <width>300</width>
-    <height>244</height>
+    <width>362</width>
+    <height>370</height>
    </rect>
   </property>
   <property name="windowTitle">
    <string>Settings</string>
   </property>
   <layout class="QVBoxLayout" name="verticalLayout">
-   <item alignment="Qt::AlignTop">
-    <widget class="QGroupBox" name="behaviourGroup">
-     <property name="title">
-      <string>Behaviour</string>
+   <item>
+    <widget class="QTabWidget" name="tabWidget">
+     <property name="currentIndex">
+      <number>0</number>
      </property>
-     <layout class="QVBoxLayout" name="verticalLayout_2">
-      <item alignment="Qt::AlignTop">
-       <widget class="QCheckBox" name="translateImmediatelyCheckbox">
-        <property name="text">
-         <string>Translate as I type</string>
-        </property>
-       </widget>
-      </item>
-     </layout>
-    </widget>
-   </item>
-   <item alignment="Qt::AlignTop">
-    <widget class="QGroupBox" name="resourceGroup">
-     <property name="title">
-      <string>Resource usage</string>
-     </property>
-     <layout class="QFormLayout" name="formLayout_2">
-      <item row="0" column="0">
-       <widget class="QLabel" name="coresLbl">
-        <property name="text">
-         <string>Threads</string>
-        </property>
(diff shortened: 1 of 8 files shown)


XapaJIaMnu:
I compiled and tried.
1) We should do something about auto-resizing the widget when switching to the languages tab, as the user needs to resize things to see the full name.
2) We should give the user the information why some models are undeleteable (eg, they were distributed with the exe, and the user can just remove them from the current directory)

In terms of validation, we this is already done by this function: 
https://github.com/XapaJIaMnu/translateLocally/blob/db88f0d571055ae7750a72f35413575e2dad6f6f/src/ModelManager.cpp#L144
Difference being, it takes a parent folder. Maybe some of it can be abstracted to work on a specific model file. (OR you know use this function: https://github.com/XapaJIaMnu/translateLocally/blob/db88f0d571055ae7750a72f35413575e2dad6f6f/src/ModelManager.cpp#L37 )

As a side note, we probably should explain in more details the change dir shenanigans that go on in here: https://github.com/XapaJIaMnu/translateLocally/blob/db88f0d571055ae7750a72f35413575e2dad6f6f/src/ModelManager.cpp#L189

src/TranslatorSettingsDialog.ui
```diff
@@ -6,63 +6,155 @@
    <rect>
     <x>0</x>
     <y>0</y>
-    <width>300</width>
-    <height>244</height>
+    <width>362</width>
```
XapaJIaMnu, line 9: Every time one of us works on QtCreator, we have phantom resizes here and there. Not sure how we can completely avoid that.

jelmervdl: Probably by being a bit more strict in which lines we commit. This time it was intentional though, 300px was a bit too small for the dialog to also contain the tab pane and the table. 362 is a bit random…

XapaJIaMnu: Not a criticism, just an observation. I've noticed this with my commits as well.

jelmervdl:
A problem with the current approach to models is that we really trust the .tar.gz files.

We assume
- that the name of the .tar.gz file is also the name of the folder it contains
- that it contains a single folder with all the model files
- that no two different models shall have the same folder name
- that it doesn't contain any symlinks or paths with ".." in it (I've seen that libarchive has some flags to guarantee this, but we don't explicitly set them)

Additionally, right now if you upgrade a model it does so by extracting over the current model. This could leave files from the old model that haven't been overwritten lying around. It also corrupts the old model when the updated model fails to extract completely.

I think we could solve all of the above if:
- We always extract models to a unique folder (say like UUID based folder names)
- If all entries in the .tar.gz indeed have the same prefix folder, we substitute that for the unique folder. Otherwise we extract all files as is to the unique folder (so models can still have subdirectories if they need to)
- If extraction fails half-way through, or the model_info.json file is not found, we can just delete the unique folder
- If it does extract & loads correctly, and it is the same model (same long name?) but a newer version of an already installed model, we can delete the old model completely by just removing that unique folder. It won't affect the updated model.

jelmervdl:
Additionally we might want to add checksum to the model index at some point? Then we can just point to models on S3 or 3rd party sites and validate the .tar.gz files before extracting them. Won't solve the above issues for locally imported models (e.g. by using the "Import models…" button I added.

XapaJIaMnu:
I'm all for md5 checksums being distributed alongside the tar.gz possibly inside the models.json which we fetch from the server. Then we can download and compute the md5sum.

Can we not do opaque folder names? We could add a timestamp to them, I guess. I see little benefit of it though, we could just delete the previous folder before extracting?

kpu:
md5 is broken.  Let's do something more secure.

jelmervdl:
Any of these would be fine with me: https://doc.qt.io/qt-5/qcryptographichash.html#Algorithm-enum

jelmervdl:
> Can we not do opaque folder names?

I don't trust the archives to always have a folder with a unique name in them, or a consistent name across upgrades. Like, that would be okay if it was just a single source for models. But I just added a button to import your own .tar.gz files, opening it up for all kinds of shenanigans.


> we could just delete the previous folder before extracting?

I'm thinking of the case extraction might fail, or it might be an incomplete archive. Then we also just deleted the old working version.

> We could add a timestamp to 
Them

That would be sufficient to get around the above failure case

XapaJIaMnu:
We are the single source models for now, but yeah, we might not be in the future. I don't see users being too keen to prepare their own models, but hopefully that will change. And when they do, they would expect models with the same folder names to override the previous model they tried to import. I guess confirmation for "Folder X, exists, replace?" would work here.
We also have to teach people how to prepare their `model_info.json`, I guess.

I'm not sure adding an extra timestamp to the folder name is absolutely necessary. It is defnitely not harmful.

XapaJIaMnu:
Delete models was somewhat buggy for me. I had 5 models and I deleted them one by one and at first, multiple models disappeared at once, then later on I kept having 1 model left, but as I deleted it, another one appeared. Could it because I had the .tar.gz files inside the directory, as per the behaviour of the master branch?

The initial model manager window is a bit tiny, especially if one has many models installed.

XapaJIaMnu:
Same run, when I try to download a new model after deleting everything:
Screenshot_20210624_183123

jelmervdl:
Oooh interesting! Why would it not be able to move the folder? Hmm… I'll experiment a bit more.

I'll also look into the weird deleting/popping back models. Probably a signalling issue in ModelManager. Making the QAbstractTableModel emit the right signals is a bit tricky.

XapaJIaMnu:
Same issue, if I try to import a model. I assume it works for you on mac?

jelmervdl:
Yeah, worked fine for me. Not sure why it doesn't work for linux. Does the `~/.config/translateLocally` path exist when you encounter this error?

XapaJIaMnu:
Yes, it does exist. It is created automatically on the first run if it doesn't exist.

jelmervdl:
I have no clue why it would not just work. QDir::rename() and QTemporaryDir both don't do anything magical. I'll try it in my Ubuntu VM.

jelmervdl:
I haven't been able to replicate the rename() failure. What OS are trying this on (looks like something with KDE)

Do you have any special setup around your temp and home directory? Like, could there be any reason why a directory in /tmp could not be moved to ~/.config/translateLocally? Different filesystem maybe?

Lastly, could you try the commented out line in ModelManger.cpp that creates a temp directory in the translateLocally directory directly, and see if that solves it? I.e. change ModelManager.cpp:91 to:
```cpp
// Initially extract to to a temporary directory. Will delete its contents
// when it goes out of scope
QTemporaryDir tempDir(configDir_.filePath(".extracting-XXXXXXX"));
// QTemporaryDir tempDir;
```

XapaJIaMnu:
I'll take a look over the weekend. I suspect it might be an issue with the working directory not being clean due to it being created by a previous version of translateLocally. 

I use arch Linux

XapaJIaMnu:
I think i found the issue. Across SO:
> ...
> Do remember that even with giving a full path QDir::rename will fail if source and destination path's are not on the same partition or for a few more reasons as mentioned in the documentation.
> ...
My tmp is not on the same partition.

XapaJIaMnu:
This will potentially be a problem on a number of systems that use different drive configurations. With some brief googling, it seems everyone recommends the use of `QDir::rename` which will fail in my case. Not sure what's the next best thing to do. Maintain our own `temp` in the configuration dir? Use C++17 filesystem?

jelmervdl:
I think the safest work-around is just creating a temp directory inside the config dir. This also saves us from creating temp files and then ending up trying to move them to a filesystem that's full or has a quota on it that causes the move to fail.

I've explicitly excluded any of the temp directories from being scanned. Not that they should still exist at that point, but who knows what weird crashes or bugs may cause them to survive their pruning. I thought about just making them hidden directories and excluding them that way from the scan in `scanForModels()`, but I'm not sure that works on Windows.

XapaJIaMnu:
Yes, this finally works for me.

How about deleting any `extracting-XXXX` directories when the program is run? That should do a proper cleanup in case we had a crash during model extraction on a previous run.

XapaJIaMnu:
Or is this done automatically by the temporary dir function?

jelmervdl:
Yes. the QTemporaryDir will delete the directory it made (recursively) when it's deallocated. It's only when the application segfaults (e.g. somewhere in libarchive) that the `extracting-XXXX` folder would stay.

I'm not a fan of the idea of deleting these folders on start-up. I think Linux and Windows there's nothing preventing you from starting a second instance of the program. If you would do that exactly while another instance is extracting, it would break that process.

XapaJIaMnu:
Fair point, merging.

Issue #21, opened 2021-05-12
Security


kpu:
Models appear to be downloaded over HTTP without some form of validation.  
http://data.statmt.org/bergamot/models/models.json
One can overwrite arbitrary files with a Marian configuration.  
We could ship checksums as metadata or enable HTTPS properly.

XapaJIaMnu:
I can add a checksum to the json, I guess. Why does data.statmt.org not have https version?

jelmervdl:
Would it make sense to host the models.json file somewhere on Github? That would give us stable hosting + https + some way to allow others to contribute models for pretty much free.

If there are checksums for the downloads in the json file, it doesn't matter much where those models come from. They can stay on http-only data.statmt.org or some CDN if they ever become annoying to host.

kpu:
I can host the JSON on https://translatelocally.com/

XapaJIaMnu:
None whatsoever. I don't have access to whoever hosts translateLocally.com so I someone has to manually copy that from the students deployment repo.

kpu:
Updated.

XapaJIaMnu:
The models.json is generated by this script: https://github.com/browsermt/students/blob/master/generate_models_json.py which is called repeatedly by https://github.com/browsermt/students/blob/master/deploy.students.sh
We should incorporate the automatic sha256sum field in there.

XapaJIaMnu:
While this is merged now, I'd like to keep this open until the student repo automatically generates the checksums.

jelmervdl:
I'll make a pull request on the students repo that adds the sha256 checksums.

jelmervdl:
Forgot to tag this issue in the pull request. It's here: https://github.com/browsermt/students/pull/39

XapaJIaMnu:
Fixed by @jelmervdl

PR #49, opened 2021-06-25
Use https + checksums for model security

jelmervdl:
- This change switches the model list url to https://translatelocally.com/models.json for the https as per #21.
- Adds a checksum to validate the download, as discussed in #48.

This change adds support for an optional `checksum` property in the model list JSON, which is a hexadecimal sha256 checksum of the file. If this checksum is present, and it does not match that of the downloaded file, the downloaded archive is not extracted and thrown away.

I've used sha256 as that seems to be pretty commonly available. macOS, Windows and Linux ship with command line tools for generating these checksums.

### DIFF
--- a/src/Network.cpp
+++ b/src/Network.cpp
@@ -19,20 +19,38 @@ QNetworkReply* Network::get(QNetworkRequest request) {
     return nam_->get(request);
 }
 
-QNetworkReply* Network::downloadFile(QUrl url, QFile *dest) {
+QNetworkReply* Network::downloadFile(QUrl url, QFile *dest, QCryptographicHash::Algorithm algorithm, QByteArray hash) {
     QNetworkReply *reply = get(QNetworkRequest(url));
-    
+
     // Open in read/write so we can easily read the data when handling the
     // downloadComplete signal.
     if (!dest->open(QIODevice::ReadWrite)) {
         emit error(tr("Cannot open file for downloading."));
         return nullptr;
     }
+
+    // Optionally we have a hasher that hashes the download as it comes in and
+    // compares it against the provided hash when the download completes.
+    QCryptographicHash *hasher = nullptr;
+    if (!hash.isEmpty()) {
+        hasher = new QCryptographicHash(algorithm);
+
+        // Free the hasher when the reply is destroyed. All methods that access the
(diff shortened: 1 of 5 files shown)


src/Network.h
```diff
@@ -19,19 +20,19 @@ class Network : public QObject {
 
     /**
      * Download a file to a destination on disk. Useful for skipping loading the
-     * file in memory. The `downloadComplete(QFile,QString)` signal will be 
-     * emitted with a pointer to the file and a suggested filename. During 
-     * download the `progressBar(qint64,qint64)` signal will be emitted.
+     * file in memory. Optionally a sha256The `downloadComplete(QFile,QString)` signal will be
```
XapaJIaMnu, line 23: Phantom `The` here?

src/Network.cpp
```diff
@@ -19,20 +19,38 @@ QNetworkReply* Network::get(QNetworkRequest request) {
...
     // Open in read/write so we can easily read the data when handling the
     // downloadComplete signal.
     if (!dest->open(QIODevice::ReadWrite)) {
         emit error(tr("Cannot open file for downloading."));
         return nullptr;
     }
+
+    // Optionally we have a hasher that hashes the download as it comes in and
+    // compares it against the provided hash when the download completes.
+    QCryptographicHash *hasher = nullptr;
```
XapaJIaMnu, line 34: Why not use a smart ptr here, as opposed to a signal for freeing later?

XapaJIaMnu (approved):
Minor point: I think we should have a smart pointer for the hasher deletion, as opposed to hooking a signal to do it later. This will improve readability.

Another minor point. I think we should compute the hash every time and only check it against the reference if the reference hash exists. This will simplify the control flow in the function a bit.

src/Network.cpp
jelmervdl: Because that's the sensible thing to do.

XapaJIaMnu: Sorry, could you elaborate? Why not do something like `std::unique_ptr<QCryptographicHash> hasher(nullptr);`

XapaJIaMnu: Oh you've already changed it.

XapaJIaMnu:
Merge when ready, unless you want to commit more fixes?

Issue #50, opened 2021-07-07
Show alignment data in translation


jelmervdl:
[bergamot-translate exposes alignment information](https://github.com/browsermt/bergamot-translator/blob/main/src/translator/response.h) and it would be cool to expose this in the interface.

I propose we initially use it to sync the visible part in the input and translated text boxes. I.e. make sure the part you're editing in the source text is also visible in the translated text box so you immediately see the results of your editing.

For example, when in Overleaf you click somewhere in the preview document, it will move the cursor in the edit document to the section you clicked. 

A next step (hidden behind a toggle) could be [a sort of focus mode](https://ia.net/writer), where the sentence you're editing is highlighted both in the input text box and the translated text box. (This would be a bit more tricky to implement I expect due to how the sentence boundaries in the input box need to be updated/estimated while you type, not only when translation is available.)

Another example could be what OPUS-CAT does is a synchronised hover effect.
![Screenshot 2021-07-07 at 13 41 43](https://user-images.githubusercontent.com/198639/124761409-ad5ae300-df29-11eb-9058-52fd0ff41054.png)
We could have something similar in one direction by highlighting the word/substrings in the translation that align with the word/substring that currently under/next to the caret in the input text box. For the other (translation -> source text) direction, maybe clicking/selecting a word/substring also selects the aligned parts in the input text box?

ChrisBurnsOneOne:
This is already the intention for the Firefox plugin version. Mozilla should be on the case producing a couple of slightly different prototypes which have been suggested over the last month or so.

jelmervdl:
Oeeh that would be interesting!

ChrisBurnsOneOne:
Unsure how much of it I can share in terms of basic concepts, but I Nikolai should already have seen some of this work. There's no real reason the plugin and desktop versions need to re-invent the wheel here.

XapaJIaMnu:
Closed via #52

PR #52, opened 2021-10-01
Draft: Alignment information

jelmervdl:
Implementation of #50

This pull requests adds a QSyntaxHighlighter to the translation output box that highlights the alignment probability of words (or sentencepiece tokens really) in the output box given the sentencepiece token currently under the cursor in the input box.

There's a toggle to turn it on/off in the *View* menu.

It also works when you select more than one word in the input box. It will then highlight all the parts that match that selection. The probability-based colour will make less sense in that case though.

Additionally, if you click on a word in the output box (the translation) it will move the cursor in the input box with the word that aligns best with it. However, it does not move the focus to the input box, so the only way to then use that moved cursor is through tapping [shift] + [tab]. I'm not yet sure about the usefulness of this feature.

image

### DIFF
--- a/src/Translation.cpp
+++ b/src/Translation.cpp
@@ -28,14 +28,14 @@ bool findWordByByteOffset(marian::bergamot::Annotation const &annotation, std::s
     return wordIdx != annotation.numWords(sentenceIdx);
 }
 
-qsizetype offsetToPosition(std::string const &text, std::size_t offset) {
+int offsetToPosition(std::string const &text, std::size_t offset) {
     if (offset > text.size())
         return -1;
 
     return QString::fromUtf8(text.data(), offset).size();
 }
 
-std::size_t positionToOffset(std::string const &text, qsizetype pos) {
+std::size_t positionToOffset(std::string const &text, int pos) {
     return QString::fromStdString(text).left(pos).toLocal8Bit().size();
 }
 
@@ -57,7 +57,7 @@ QString Translation::translation() const {
     return QString::fromStdString(response_->target.text);
 }
 
-QList<WordAlignment> Translation::alignments(qsizetype sourcePosFirst, qsizetype sourcePosLast) const {
+QList<WordAlignment> Translation::alignments(int sourcePosFirst, int sourcePosLast) const {
     QList<WordAlignment> alignments;
     std::size_t sentenceIdxFirst, sentenceIdxLast, wordIdxFirst, wordIdxLast;
 
@@ -102,7 +102,7 @@ QList<WordAlignment> Translation::alignments(qsizetype sourcePosFirst, qsizetype
     return alignments;
 }
 
-qsizetype Translation::findSourcePosition(qsizetype targetPos) const {
+int Translation::findSourcePosition(int targetPos) const {
     std::size_t sentenceIdx, wordIdx;
 
     if (!response_)
--- a/src/Translation.cpp
+++ b/src/Translation.cpp
@@ -1,0 +1,134 @@
+#include "Translation.h"
+#include "3rd_party/bergamot-translator/src/translator/response.h"
+#include <QDebug>
+
+namespace {
+
+bool contains(marian::bergamot::ByteRange const &span, std::size_t offset) {
+    // return offset >= span.begin && offset < span.end;
+
+    // Note: technically incorrect, but gives better user experience for those
+    // not aware exactly what is happening with sentence piece tokens.
+    return offset > span.begin && offset <= span.end;
+}
+
+bool findWordByByteOffset(marian::bergamot::Annotation const &annotation, std::size_t pos, std::size_t &sentenceIdx, std::size_t &wordIdx) {
+    for (sentenceIdx = 0; sentenceIdx < annotation.numSentences(); ++sentenceIdx) {
+        if (::contains(annotation.sentence(sentenceIdx), pos))
+            break;
+    }
+
+    if (sentenceIdx == annotation.numSentences())
+        return false;
+    
+    for (wordIdx = 0; wordIdx < annotation.numWords(sentenceIdx); ++wordIdx)
+        if (::contains(annotation.word(sentenceIdx, wordIdx), pos))
+            break;
+
+    return wordIdx != annotation.numWords(sentenceIdx);
+}
+
+qsizetype offsetToPosition(std::string const &text, std::size_t offset) {
+    if (offset > text.size())
+        return -1;
+
+    return QString::fromUtf8(text.data(), offset).size();
+}
+
+std::size_t positionToOffset(std::string const &text, qsizetype pos) {
+    return QString::fromStdString(text).left(pos).toLocal8Bit().size();
+}
+
+} // Anonymous namespace
+
+Translation::Translation()
+: response_()
+, speed_(-1) {
+    //
+}
+
+Translation::Translation(marian::bergamot::Response &&response, int speed)
+: response_(std::make_shared<marian::bergamot::Response>(std::move(response)))
+, speed_(speed) {
+    //
+}
+
+QString Translation::translation() const {
+    return QString::fromStdString(response_->target.text);
+}
+
+QList<WordAlignment> Translation::alignments(qsizetype sourcePosFirst, qsizetype sourcePosLast) const {
+    QList<WordAlignment> alignments;
+    std::size_t sentenceIdxFirst, sentenceIdxLast, wordIdxFirst, wordIdxLast;
+
+    if (!response_)
+        return alignments;
+
+    if (sourcePosFirst > sourcePosLast)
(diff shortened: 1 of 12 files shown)


jelmervdl:
It makes It completely breaks down once you turn off "translate as I type" and start typing: It will still highlight stuff, but that doesn't make any sense at that point.

As far as I know the QTextEdit widget doesn't give me enough information or a data type that allows me to see which parts are original, and which are added, and backtrack where the cursor position would be if we'd remove all the changes since the last translation. Therefore I'm thinking of just disabling highlighting as soon as you start changing the input text.

qianqianzhu:
> FYI, I intend to remove softAlignment/hardAlignment support in the cache PR ([browsermt/bergamot-translator#202](https://github.com/browsermt/bergamot-translator/pull/202)) most likely. I mean, it'll work given there's a commit for bergamot-translator the submodule points to for now, but whatever API for alignments provided is unlikely to survive @qianqianzhu's TagTree PR - [browsermt/bergamot-translator#199](https://github.com/browsermt/bergamot-translator/pull/199) (conditioned on the functionality working).

@jerinphilip feel free to go for your plan. Just one question: how can I get Marian alignment matrix after that?

kpu:
Please kill all hard alignments.  I think there is a usecase for soft alignments to be returned to the user if requested in raw form.  Keep in mind @qianqianzhu can be stacked on top, taking the soft alignments and tag information.

XapaJIaMnu:
I can't get it to work with QT5 on Linux. I get a single error.
```
QObject::connect: Cannot queue arguments of type 'Translation'
(Make sure 'Translation' is registered using qRegisterMetaType().)
```
No translations are displayed at this point, regardless of whether i use translate as I type or the translate button. Showing alignments also doesn't make a difference.

src/Translation.h
```diff
@@ -0,0 +1,73 @@
...
+ * interface from the rest of the Qt code, and provides utility functions to
+ * access alignment information with character offsets instead of byte offsets.
+ */
+class Translation {
+private:
+    // Note: I would have liked unique_ptr, but that does not go well with
+    // passing Translation objects through Qt signals/slots.
+    std::shared_ptr<marian::bergamot::Response> response_;
+
+    int speed_;
```
XapaJIaMnu, line 35: The comment with regards what `speed_` is got lost with the code re-organisation.

XapaJIaMnu (commented):
In order to solve the issue of being able to highlight inside long blocks of text, we probably need a more complicated container for the text, before it gets dispatched to qTextBox. Let's discuss offline.

jelmervdl:
> In order to solve the issue of being able to highlight inside long blocks of text, we probably need a more complicated container for the text, before it gets dispatched to qTextBox. Let's discuss offline.

I found https://doc.qt.io/qt-5/qtextdocument.html#contentsChange which gives information about edits and their position. With that we should be able to track and backtrack where the cursor is given the original alignment offsets. But it will involve a bit of bookkeeping.

XapaJIaMnu:
It works now. I think the alignment works quite well even with translate-as-i-type off. If you hover over the text in the input that hasn't been translated, it just points to end-of-sentence, which is kind of expected.

The behaviour is kind of strange when you try to click on the translated text. I would expect it to highlight something on the input side, not move my cursor (also, the cursor thing doesn't work on Linux it seems?

jelmervdl:
>  I think the alignment works quite well even with translate-as-i-type off.

At the end it seems to be quite okay, but when you start editing your input text in the middle of the text, any word you select after your edited parts will be misaligned.

> The behaviour is kind of strange when you try to click on the translated text. I would expect it to highlight something on the input side, not move my cursor (also, the cursor thing doesn't work on Linux it seems?

I think that's the more expected behaviour; I'll change it to basically have the same highlighting behaviour as the input box has on the output box, and then switch which box receives the alignment highlighting based on which one is focussed. See how well that would work.

XapaJIaMnu:
Set a "dirty" bit on the input text and only compute alignment after the dirty bit is off?

jelmervdl:
> Set a "dirty" bit on the input text and only compute alignment after the dirty bit is off?

Good idea. I've used the isModified/setModified flags on the input text document to track whether it's dirty now, and only show any highlight information if it isn't. It's less fancy than tracking changes and keeping up to date with the updated offsets, but it's a whole lot easier to implement!

XapaJIaMnu:
> > Set a "dirty" bit on the input text and only compute alignment after the dirty bit is off?
> 
> Good idea. I've used the isModified/setModified flags on the input text document to track whether it's dirty now, and only show any highlight information if it isn't. It's less fancy than tracking changes and keeping up to date with the updated offsets, but it's a whole lot easier to implement!

I think users could live with that.

src/Translation.cpp
```diff
@@ -0,0 +1,107 @@
...
+
+int offsetToPosition(std::string const &text, std::size_t offset) {
+    if (offset > text.size())
+        return -1;
+
+    return QString::fromUtf8(text.data(), offset).size();
+}
+
+std::size_t positionToOffset(std::string const &text, int pos) {
+    return QString::fromStdString(text).left(pos).toLocal8Bit().size();
```
XapaJIaMnu, line 37: Maybe change here to use `fromUtf8`?

XapaJIaMnu (commented):
So, from the QT documentation I read this:

> In addition to QString, Qt also provides the QByteArray class to store raw bytes and traditional 8-bit '\0'-terminated strings. For most purposes, QString is the class you want to use. It is used throughout the Qt API, and the Unicode support ensures that your applications will be easy to translate if you want to expand your application's market at some point. The two main cases where QByteArray is appropriate are when you need to store raw binary data, and when memory conservation is critical (like in embedded systems).

if we are able to work with `QByteArray`, we would have easier time?  https://doc.qt.io/qt-5/qbytearray.html#fromStdString

That should save some of the conversions and `QString` has zero copy constructor from `QByteArray`

What do you think?

jelmervdl:
Notes to self, or diagnosis of injuries currently observed in feet:
- highlighter triggers textChanged signal, which causes feedback loop when *translate as I type* is on.
- handle 0 sentences case
- `X < 0ul - 1` does not work well.
- The gaps between sentences do not appear to be parts of sentences in AnnotatedText.

src/Translation.cpp
jelmervdl: In this case I need `QString::left()` because of char pos. The `size` argument of `fromUtf8` won't do since that's bytes.

I checked whether there would be any other benefit for using `fromUtf8`, but `fromStdString` appears to be identical to `fromUtf8(text.data(), int(text.size()))`.

XapaJIaMnu: Hmm... It seems all of them do a copy of the memory ;/

jerinphilip:
> handle 0 sentences case

https://github.com/browsermt/bergamot-translator/issues/101#issuecomment-821162120 xD

> The gaps between sentences do not appear to be parts of sentences in AnnotatedText.

Is there a use-case for having this? Given EOSs are now marked explicitly, shouldn't [`sentenceAsByteRange(i).end, sentenceAsByteRange(i-1).begin)` provide you this already?

src/Translation.cpp
XapaJIaMnu: `fromRawData` seems to be the only zero-copy solution, but as we discussed it requires `QChar`...

jelmervdl: I've replaced the code with something based on https://stackoverflow.com/a/3586973.

src/Translation.cpp
```diff
@@ -0,0 +1,107 @@
...
+, speed_(speed) {
+    //
+}
+
+QString Translation::translation() const {
+    return QString::fromStdString(response_->target.text);
+}
+
+QList<WordAlignment> Translation::alignments(Direction direction, int sourcePosFirst, int sourcePosLast) const {
+    QList<WordAlignment> alignments;
```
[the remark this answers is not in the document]
XapaJIaMnu: Some light bed time reading: https://www.qt.io/blog/qlist-changes-in-qt-6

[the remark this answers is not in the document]
XapaJIaMnu: Yeah... I don't imagine they'd break it, but that sounds reasonable.

XapaJIaMnu:
```
dheart@arbiter ..i_stuff/postdoc/translateLocally/build (git)-[show-alignment] % ./translateLocally
QObject::connect: Cannot queue arguments of type 'Translation::Direction'
(Make sure 'Translation::Direction' is registered using qRegisterMetaType().)
```
One issue that I have when copy/pasting large chunks of texts is that the two textbox screens are out of sync. Eg the input one would have scrolled to the end, whereas the output one would stay at the start. I think that on click in the input field, it should scroll the output field to the corresponding place and vice versa?

PR #58, opened 2021-11-09
Add an initial command line interface

### LINKED ISSUE #51: Make it usable from the shell
I'd love to be able to `echo ikkje så gale | translateLocally nno-eng` (also handy for chaining with other engines like apertium)

XapaJIaMnu:
We had a request do to a simple command line interface. What this does is basically, implement a very basic command line interface that reuses the models and the settings downloaded via the GUI but allows for piping or daisy chaining for pivoting. This will eventually fix #51 
Example usage:
Listing the models:
```
./translateLocally -l
Czech-English type: tiny version: 1; To invoke do -m cs-en-tiny
German-English type: tiny version: 2; To invoke do -m de-en-tiny
English-Czech type: tiny version: 1; To invoke do -m en-cs-tiny
English-German type: tiny version: 2; To invoke do -m en-de-tiny
English-Spanish type: tiny version: 1; To invoke do -m en-es-tiny
```

Translating to a file:
```
./translateLocally --model en-es-tiny -i /tmp/tst -o /tmp/tran
```

Translating from stdin/stdout:
```
echo "Me gustaria comprar la casa verde" | ./translateLocally -m es-en-tiny
I'd like to buy the green house.
```

Pivoting:
```
echo "Me gustaria comprar la casa verde" | ./translateLocally -m es-en-tiny | ./translateLocally -m en-de-tiny
Ich möchte das Gewächshaus kaufen.
```

Major issues.
Batching. Since I didn't want to re-implement marian-decoder with batching, or modify the `marianInterface` in any ways, so I just do a `readAll()`, which won't scale to big files. Maybe users won't use it for big files.  Maybe we should use the service more directly?
Blocking. I'm using a `QEventLoop` to wait for the translation to complete. Maybe we should really not be using the service here..?
Fun thing how a house that is green in Spanish becomes Greenhouse in German...

### DIFF
--- a/src/CommandLineIface.cpp
+++ b/src/CommandLineIface.cpp
@@ -1,0 +1,90 @@
+#include "CommandLineIface.h"
+#include <QFile>
+
+
+CommandLineIface::CommandLineIface(QCommandLineParser& parser, QObject * parent) : QObject(parent), eventLoop_(this), parser_(parser), models_(this),
+                                                                                   settings_(this), translator_(new MarianInterface(this)),
+                                                                                   qcin_(stdin), qcout_(stdout), qcerr_(stderr) {
+    connect(translator_, &MarianInterface::error, this, &CommandLineIface::outputError);
+    connect(translator_, &MarianInterface::translationReady, this, &CommandLineIface::outputTranslation);
+}
+
+int CommandLineIface::run() {
+    if (parser_.isSet("l")) {
+        printLocalModels();
+        return 0;
+    } else if (parser_.isSet("m")) {
+        QString model_shortname = parser_.value("model");
+
+        // Try to find our model in the list of models
+        QString modelpath("");
+        for (auto&& model : models_.getInstalledModels()) {
+            if (model.shortName == model_shortname) {
+                modelpath = model.path;
+            }
+        }
+        if (modelpath == "") {
+            qcerr_ << "We could not find a model identified as: " << model_shortname << " . Use translateLocally -l to list available models or use the GUI to download some from the internet.\n";
+            return 1;
+        }
+        // Init the translation model
+        translator_->setModel(modelpath, settings_.marianSettings());
+        doTranslation();
+        return 0;
+    } else {
+        qcerr_ << "We are in command line mode, but there's nothing for us to do. Some control flow mistake maybe?\n";
+        return 2;
+    }
+}
+
+void CommandLineIface::printLocalModels() {
+    QList<Model> localmodels = models_.getInstalledModels();
+    for (auto&& model : localmodels) {
+        qcout_ << model.src << "-" << model.trg << " type: " << model.type << " version: " << model.localversion << "; To invoke do -m " << model.shortName << '\n';
+    }
+}
+
+/* This function is pseudo blocking, via an event loop.*/
+void CommandLineIface::doTranslation() {
+    // Find whether input is stdin or a file
+    QString input;
+    if (parser_.isSet("s")) {
+        QFile inputFile(parser_.value("s"));
+        if (inputFile.open(QIODevice::ReadOnly)) {
+           QTextStream in(&inputFile);
+           input = in.readAll(); // @TODO this should be more complicated batching thingie, but alas. Should work fine with most small inputs.
+                                 // Alternative is to read it line by line but then it would be horribly inefficient as we would lose all batching.
+           inputFile.close();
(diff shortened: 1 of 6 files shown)


src/main.cpp
```diff
@@ -22,13 +23,18 @@ int main(int argc, char *argv[])
...
 
-    // Launch application
-    MainWindow w;
-    w.show();
-    return translateLocally.exec();
+    // Launch application unless we're supposed to be in CLI mode
+    if (translateLocally::isCLIOnly(parser)) {
+        // The CLI is supposed to just do something and then die.
+        // To do that we will just have an object whose constructor as a side
+        // effect will do the thing and die.
```
jelmervdl, line 32: Is this comment still true? Looks like the meaty part of the CLI is implemented implemented in `run()`, not the constructor. There don't seem to be all that many side effects to the constructor :D

src/CommandLineIface.cpp
```diff
@@ -0,0 +1,90 @@
...
+
+void CommandLineIface::outputTranslation(Translation output) {
+    // Find whether output is stdin or a file
+    if (parser_.isSet("o")) {
+        QFile outputFile(parser_.value("o"));
+        if (outputFile.open(QIODevice::WriteOnly)) {
+            QTextStream out(&outputFile);
+            out << output.translation();
+        } else {
+            outputError(QString("Couldnl't open output file: " + parser_.value("o")));
```
jelmervdl, line 84: Typo in `couldnl't`

src/CommandLineIface.cpp
```diff
@@ -0,0 +1,90 @@
...
+           QTextStream in(&inputFile);
+           input = in.readAll(); // @TODO this should be more complicated batching thingie, but alas. Should work fine with most small inputs.
+                                 // Alternative is to read it line by line but then it would be horribly inefficient as we would lose all batching.
+           inputFile.close();
+        } else {
+            outputError(QString("Couldn't open input file: " + parser_.value("i")));
+        }
+    } else {
+        input = qcin_.readAll(); // @TODO again this should do some batching and allow for full cmd usage but I *really* don't want to re-implement marian...
+                                 // Maybe expose the raw marian interface with their batcher, or the service which has some sort of CLI usage.
```
jelmervdl, line 63: I'm tempted to actively push that kind of use case to either Marian itself, or bergamot-translator's (upcoming?) Python module.

jelmervdl (approved):

src/main.cpp
XapaJIaMnu: Yeah that's from earlier version when my design was more appalling.

src/CommandLineIface.cpp
XapaJIaMnu: Do you mean the whole CMD interface or proper batching? I should be able to write a short batcher with not too much difficulty, but I don't want to recreate functionality

jelmervdl: (In case it was a bit lost in Gather.town today) My idea is to have translateLocally be able to do command line translation, but not be best-in-class in it. If they want speed and efficiency, they should go to Marian. If they want to embed it, either marian-server or the Python interface of bergamot. If you they want to translate a file, then, okay. But I am assuming those files are a couple of MBs at most. Not GBs.

However, if a simple loop (i.e. `while file not empty; read 1000 lines, translate (with batching inside those 1000 lines handled by bergamot transparently) print translations`) can already prevent people from swamping their memory and get what they want, that would be a worthy addition. *But the goal should not be translate very large files as fast as possible, just translate large files without running out of memory*.

Additionally, ideally the translatelocally.com should have a little section on alternatives (such as marian-decoder and bergamot's Python API) for advanced use cases, hidden away for those who search for it.

XapaJIaMnu:
That should be more memory efficient, but users can't really use stdin, because we're waiting for 320 lines (or EOF), before dispatching a translation. The next thing to do is to get a some sort of timer so at readline so that we can dispatch a translation (maybe similar to the way you do it with the service).

Maybe signal and slot based `fetchData` and a input buffer solution...

XapaJIaMnu:
The proper way to do this would be to use [QSocketNotifier](https://stackoverflow.com/questions/9741259/reading-asynchronously-from-stdin-with-qt)... The only issue is that this code is not portable, as windows requires different technique. I think I will just add an option for separate `interactive` mode and call it a day.

PR #60, opened 2021-11-16
Clean-up of cli code path

jelmervdl:
While reviewing #58 I noticed that it didn't work well with files with blank lines, which might have been due to the `peek()` call in the `fetchData` method. While digging through documentation to figure out what was going on, I saw a couple of opportunities for improvements.

That if-statement `qcin_.device()->peek(inbuff.data(), 8) == 0` was true after each line on macOS when piping another process into translateLocally, which in term caused `fetchData` to only return single lines. If one of those lines was empty, it returned `\n` which caused translateLocally to stop translating.

Test case:
```
echo -ne "This is a test.\n\nWith an empty line." | ./translateLocally -m en-de-tiny
```

While debugging this (and not hitting the "submit review" button before #58 was merged, sorry!) I also noticed something else that worried me in the `QIODevice::peek()` documentation:
> Note that on sequential devices, data may not be immediately available, which may result in a partial line being returned. By calling the canReadLine() function before reading, you can check whether a complete line (including the newline character) can be read.

So digging further through the documentation, no such warning exists for `QTextStream`, which is already used for file handling. I rewrite the stdin/stdout codepath to use those same streams (it seems you can just alter where they point through after construction using `QTextStream::setDevice()`, ideal!).

While eliminating pointers, I did the same thing for the QFile pointers. The destructor of QFile also guarantees to close them on destruct, so that's cool. I also moved the `exit()` calls from the constructor to `run()` because quitting half-way through a constructor feels scary (although, it's `exit()`, it just kills the program, no proper deconstruction anyway, I suppose)

And then the game was on. So I rewrite `fetchData` to re-use its strings as much as possible, getting the translation time for Crime & Punishment from 42s to 32s when using the `-i` and `-o` options.

### DIFF
--- a/src/CommandLineIface.cpp
+++ b/src/CommandLineIface.cpp
@@ -3,10 +3,9 @@
 
 #include <array>
 
-CommandLineIface::CommandLineIface(QCommandLineParser& parser, QObject * parent)
+CommandLineIface::CommandLineIface(QObject * parent)
 : QObject(parent)
 , eventLoop_(this)
-, parser_(parser)
 , models_(this)
 , settings_(this)
 , translator_(new MarianInterface(this))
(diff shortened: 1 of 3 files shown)


XapaJIaMnu:
Sorry, i was a bit trigger-happy with merging.

I like the lack of code duplication and I was going to reuse the buffer after i got it to work but i just forgot about it. I will test my collection of edge cases and I will merge after.

XapaJIaMnu (approved):
yeah seems to work and better design than mine.

PR #61, opened 2021-11-16
Various fixes related to best QT practises

XapaJIaMnu:
Hi,

So I followed various QT best practises and fixed couple of things:

Using `arg(A,B)` as opposed to `arg(A).arg(B)`

Making const methods return references for performance reasons  (copy constructor is always invoked) and to prevent C++11 iterators from accidentally creating a const copy of the object to iterate: https://stackoverflow.com/questions/35811053/using-c11-range-based-for-loop-correctly-in-qt https://bugreports.qt.io/browse/QTCREATORBUG-25322

Providing `this` to several lambda based `connect` calls because capturing by reference would mean they might get called when the parent object was destroyed. https://stackoverflow.com/questions/27952902/qobject-context-in-qobjectconnect-function I hope this one will fix the QT warning about null receiver that gets called every once in a while.

Also, removing unused variables and changing some QLists to QVector for QT5 best practises reasons.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -193,7 +193,7 @@ set(LINK_LIBRARIES
     ${LibArchive_LIBRARIES}
     ${CMAKE_DL_LIBS}) # On unix necessary sometimes
 
-# Reserver the translateLocally name for the MacOS and Windows (potentially) executables. Rename the unix executable to that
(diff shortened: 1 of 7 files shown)


jelmervdl (approved):
LGTM Thanks!

Issue #38, opened 2021-05-30
First run users should get a popup that explains what the program does


XapaJIaMnu:
What the title says. Especially about downloading models.

jelmervdl:
Quick mock-up. Something like this?
![image](https://user-images.githubusercontent.com/198639/120205993-8a4e5c80-c222-11eb-8ddf-1ba49d053b0c.png)

Could be made a bit interactive, e.g. update the screenshot with arrows when you hover over each of the steps.

I'm a bit worried people will try to interact with the screenshot. Maybe it's technically possible to temporarily draw the main window inside that dialog while the dialog is open (and make it stand-alone as soon as you click "Close") but that also feels gimmicky. Maybe it works better without a screenshot.

XapaJIaMnu:
Hmm... I have a couple of problems with the screenshot, but I am not sure I can propose a meaningful solution.
1) The screenshot may work very differently from your platform's appearance.
2) Inevitably as you said, someone will try to interact with it, and be disappointed.

I was just thinking of a simple Dialog option such with things such as. "Go to the drop down menu to select/download translation models, go to the settings to increase/decrease the font size and to configure advanced translator settings."

This is not a very good tutorial, but may do what is necessary. I find the interface self explanatory, but a few people asked for pointers ;/

The ideal solution would be a guided, interactive wizard that expected the user to click in a few places. eg "Click the download models drop down menu" and the drop down menu is highlighted at the same time.

jelmervdl:
The other option I had in mind was something like https://introjs.com. I personally hate these, but they do guide the user quite well. And since it uses the actual application, no issues with screenshots.

There might be a qt library made by someone to implement such a tour. I don't think it's too difficult to implement ourselves, but it will be a bit more involved than just popup dialog.

XapaJIaMnu:
Yeah I like that bit, but unsure of how difficult it is, and what tools we have for implementing it.

jelmervdl:
@kpu do you mean we should ask the user on first time start up what the "Download models…" option should do? (I.e. show the list or open a website.) And second, should we after that choice immediately do that?

Very bad mock-up (problem is I wanted to make the preferred option the biggest one instead of just "yes")
![image](https://user-images.githubusercontent.com/198639/120214146-fda89c00-c22b-11eb-82ff-69f6068bb9d4.png)

XapaJIaMnu:
[...]

Issue #57, opened 2021-11-04
Distribute models with their own extension & mime-type


jelmervdl:
From #38:

> For windows at least it's better to manufacture our own file extension so the application can be launched for that extension.

I love the idea of being able to double-click a blob you've downloaded, and translateLocally just importing & selecting it.

To implement this, I think we'll need to do the following:

1.  Come up with an extension.
    - .translationmodel
    - .languagemodel
    - .tlmodel
    - .tllanguage
    - _something unique, not too long, possibly understandable_
2. Per platform…
    1. Add proper metadata to tell OS that TranslateLocally can open these files
    2. Implement message handlers that trigger when the OS wants to open a file ([macOS](https://doc.qt.io/qt-5/qfileopenevent.html), don't know about win & linux at the moment)

Then, we'll need to decide what to do when someone does open a model file:

1. As a macOS user, I would expect that double-clicking a model file will import it into my local library of models, and probably switch the interface to use that newly imported model. TranslateLocally should come to the foreground when this happens, and probably also give some confirmation that it completed. (This is how apps like iTunes and various apps with plugins behave.)
2. Another possible behaviour might be to just load the model once, but at next start-up not know about it anymore. As if the model is a ephemeral file. (This is more how something like Apple Preview would behave if you open a PDF).

I think the first makes more sense for downloaded models.

For distributing, best solution would probably be to host the files with the extension. Short-term work-around would be to allow people to download models from the website (we need to make that an option anyway) and use [the a[download] attribute](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/a#attr-download) to force it to have the right extension.

XapaJIaMnu:
We also want to distribute repositories as a mime type, moving this here from #84

PR #63, opened 2021-11-23
Be more friendly to first run

XapaJIaMnu:
This is a stopgap solution to #38 so that the users don't get the ugly error message. Not ideal, but should get the job done until we have a better looking solution.
Screenshot_20211123_153648

If the user clicks "ok", a remote model list is fetched and the drop down menu is opened. Nothing happens on Cancel.
@kpu good enough?

### DIFF
--- a/src/mainwindow.cpp
+++ b/src/mainwindow.cpp
@@ -250,7 +250,7 @@ void MainWindow::showEvent(QShowEvent *ev) {
 void MainWindow::showFirstRunHelper() {
     if (models_.getInstalledModels().isEmpty()) {
         QMessageBox firstRun(QMessageBox::NoIcon, tr("First run"),
-                               tr("Wellcome to translateLocally, our privacy focussed machine translation service!\n\n\
+                               tr("Wellcome to translateLocally, our privacy focussed machine translation system!\n\n\
 It looks like you don't have any translation models currently installed.\n\n\
 Would you like to connect to the Internet to see a list of available translation models?"),
                                QMessageBox::Ok | QMessageBox::Cancel, this);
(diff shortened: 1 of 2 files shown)


jelmervdl:
Nice! I was just working on the popup I proposed in #38 but much prefer to initially implement your solution, and then only deploy a full fix for #38 once we also have the "downloading models yourself" path figured out (i.e. #57)

I'll pull & review in a second, see whether I can break it :)

XapaJIaMnu:
Yeah, just fixed it...

XapaJIaMnu:
Let me know if you have any objections before I merge.

kpu:
I think it can be phrased better; see examples in #38.  

Welcome to translateLocally...

Currently it reads like an error message not a welcome message.

XapaJIaMnu:
Screenshot_20211124_103427
Better?

kpu:
s/service/system/

src/mainwindow.cpp
```diff
@@ -236,6 +237,30 @@ MainWindow::MainWindow(QWidget *parent)
...
+    QMetaObject::invokeMethod(this, "showFirstRunHelper", Qt::ConnectionType::QueuedConnection);
+#else
+    QMetaObject::invokeMethod(this, &MainWindow::showFirstRunHelper, Qt::ConnectionType::QueuedConnection);
+#endif
+}
+
+void MainWindow::showFirstRunHelper() {
+    if (models_.getInstalledModels().isEmpty()) {
+        QMessageBox firstRun(QMessageBox::NoIcon, tr("First run"),
+                               tr("Wellcome to translateLocally, our privacy focussed machine translation system!\n\n\
```
jelmervdl, line 253: Wellcome with double LL?

XapaJIaMnu: English is hard

PR #68, opened 2021-12-01
Fix bundle signing (properly) & other build/CI improvements

jelmervdl:
This fixes the CI and my Cmake skills

### DIFF
--- a/.github/workflows/macos-all.yml
+++ b/.github/workflows/macos-all.yml
@@ -11,6 +11,7 @@ on:
 env:
   # Customize the CMake build type here (Release, Debug, RelWithDebInfo, etc.)
   BUILD_TYPE: Release
+  qt_version: "6.2.1"
   ccache_basedir: ${{ github.workspace }}
   ccache_dir: "${{ github.workspace }}/.ccache"
   ccache_compilercheck: content
@@ -42,7 +43,21 @@ jobs:
     - name: Dependencies  # This qt6 build and we suppose apple accelerate was installed, but we can use openblas as a fallback until this is merged.
       run: | 
            brew update
-           brew install protobuf coreutils qt libarchive openblas ccache
+           brew install protobuf coreutils libarchive openblas ccache
+
+    - name: Cache Qt
+      id: cache-qt
+      uses: actions/cache@v2
+      with:
+        path: "${{ github.workspace }}/qt/"
+        key: qt-${{ runner.os }}-${{ env.qt_version }}
+
+    - name: Install Qt
+      uses: jurplel/install-qt-action@v2
+      with:
+        cached: ${{ steps.cache-qt.outputs.cache-hit }}
+        version: ${{ env.qt_version }}
+        dir: "${{ github.workspace }}/qt/"
     
     # Openblas location is exported explicitly because openblas is keg-only,
     # which means it was not symlinked into /usr/local/.
@@ -87,10 +102,6 @@ jobs:
       # We'll use this as our working directory for all subsequent commands
       run: cmake -E make_directory ${{github.workspace}}/build
 
-    - name: "Fix for macdeployqt (see #69)"
-      shell: bash
-      run: ln -s $(brew --prefix qt)/lib ${{github.workspace}}/build/lib
-
     - name: Configure CMake
       # Use a bash shell so we can use the same syntax for environment variable
       # access regardless of the host operating system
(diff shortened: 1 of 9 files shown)


.github/workflows/macos-all.yml
```diff
@@ -1,10 +1,24 @@
 name: macOS-{10.15-11.0}
 
-on: [push, pull_request]
```
XapaJIaMnu, line 3: Very neat, that means, CI is run every time you push to master and every time you push to a PR that is against master, correct?

jelmervdl: I think it's every push to master and every push to a pull request (or when you open a PR it will also start building that last commit to that branch)

.github/workflows/macos-all.yml
```diff
@@ -55,26 +103,55 @@ jobs:
...
       working-directory: ${{github.workspace}}/build
       shell: bash
       run: |
+          echo "arch=`../dist/cpu_feature.sh`" >> $GITHUB_ENV
           mv translateLocally.dmg translateLocally.`../dist/cpu_feature.sh`.dmg
           ls *.dmg
 
+    - uses: actions/upload-artifact@v2
+      with:
+        name: translateLocally.latest.${{ matrix.os }}.${{ env.arch }}.dmg
```
XapaJIaMnu, line 152: I think this takes care of the name. We'll see once we release.

CMakeLists.txt
```diff
@@ -250,13 +255,16 @@ if(APPLE) # Apple specific installation of the app
...
     MACOSX_BUNDLE_SHORT_VERSION_STRING ${PROJECT_VERSION_MAJOR}.${PROJECT_VERSION_MINOR}.${PROJECT_VERSION_PATCH}
+    MACOSX_BUNDLE_GUI_IDENTIFIER "com.translatelocally.translateLocally"
     )
   # When on MacOS, we have a non-standard libarchive location because we get it via brew
   target_include_directories(translateLocally-bin PRIVATE ${LibArchive_INCLUDE_DIR})
   target_include_directories(translateLocally PRIVATE ${LibArchive_INCLUDE_DIR})
-  # produce .dmg
+  # produce .dmg This requires custom target due to https://gitlab.kitware.com/cmake/cmake/-/issues/16270
+  # use dist/macdmg.sh to sign and notarize
   include(macdeployqt)
```
XapaJIaMnu, line 268: I think we should update README mac deployment instructions, since now it's no longer `cmake .. && make`, right?

jelmervdl: You're right, it would be kind of us to update the readme. Now it's either `make translateLocally && ../dist/macdmg.sh translateLocally.app` (with the right env variables, probably not worth documenting in detail) or `make translateLocally.dmg` (for your self-signed build) right?

jelmervdl:
I just noticed that sandboxing + the CLI options for input files don't work well together. I'll search a bit to see whether there is a proper way of solving this, but that's a bug related to this pull request at the moment I suppose.

XapaJIaMnu:
@jelmervdl they can still use stdin/stdout. We can  write this in the limitation and if anyone wants files, they can do it through translatelocally-bin?

jelmervdl:
I've added a hint to the output when an error occurs that could have been caused by the sandbox. I've also mentioned it in the readme.

I didn't document how to compile on macOS yet in README because I couldn't easily figure out how to install Intel MKL? Has it been renamed or something? It should probably also note that if you want to make a version that can be shared, you should use Qt's own distribution, and not homebrew. You still need homebrew to get libarchive though 🤔

XapaJIaMnu:
On MacOs we don't use MKL. We use apple accelerate framework by default when we detect MacOS, as it is installed practically everywhere.

src/CommandLineIface.cpp
```diff
@@ -1,12 +1,37 @@
...
+namespace {
+    void checkAppleSandbox(QCommandLineParser const &parser) {
+        QProcessEnvironment env(QProcessEnvironment::systemEnvironment());
+        if (!env.contains("APP_SANDBOX_CONTAINER_ID"))
+            return;
+
+        QString command("translateLocally");
+
+        if (parser.isSet("m"))
+            command += QString(" -m %1").arg(parser.value("m"));
```
XapaJIaMnu, line 20: Why is the model affected by the sandboxing? Or, alternatively, why are other options such as listing, downloading and deleting not included here?

jelmervdl: This bit is only called when `-i` or `-o` are used and cause an error. Since both these options require `-m` to be specified I found it would make sense to include `-m` in the suggested alternative command as well.

PR #71, opened 2021-12-13
Merge ubuntu & mac workflows so they can merge into one release

jelmervdl:
One big experiment™

I can't find a way to make a job (the one that makes the Github release) to depend on steps from multiple workflows. However our workflows are just one job each, so combining them should be easy and without much consequences.

This pull request tests that hypothesis

(If only Github provided a way to test github actions less publicly…)


XapaJIaMnu:
I was thinking of using an external server to scp the models and use a cron job that check a directory for new files every minute and does the release, but this would be better, indeed, thank you!

XapaJIaMnu (approved):
Thanks!

PR #72, opened 2021-12-14
Build avx and avx2 by default

XapaJIaMnu:
@jelmervdl this should take care of our arch woes, provide we merge with upstream bergamot-translator and https://github.com/browsermt/bergamot-translator/pull/275


XapaJIaMnu:
Current state. Build succeeds but it's a dynamic build with regards to qt libraries, which is not acceptable.

XapaJIaMnu:
WIndows build now succeeds statically ( i will test at some point) but I have broken vcpkg cache (which used to work). Not sure exactly how to fix it, but trying. Also probably should cache not the directory but the installed packages, so that when github updates a package of the pre-installed vcpkg we can rebuild.

jelmervdl:
I found https://github.com/microsoft/vcpkg/blob/master/docs/users/binarycaching.md#github-packages which seems to be an "official" way of caching binaries with vcpkg?

It does seem to indicate that you'd also need to install your own copy of vcpkg itself as the one Github provides is outdated.

But if the current cache action works, I'd be happy to see this merged. Compiling windows binaries is something that I haven't been able to accomplish locally. This would really help testing on that platform as well.

.github/workflows/build.yml
```diff
@@ -253,13 +271,108 @@ jobs:
...
+      run: |
+        cd installed/x64-windows-static/lib
+        # the ssplit-cpp finds staticd libs only because the extension is hardcoded to only find the debug
+        # version of the static lib: https://github.com/browsermt/ssplit-cpp/blob/master/cmake/FindPCRE2.cmake#L66
+        # or it could be a number of other reasons. Luckily we have the power of cp to rule the universe.
+        cp pcre2-8-static.lib pcre2-8.lib
+        cp pcre2-8-static.lib pcre2.lib
+        cp pcre2-16-static.lib pcre2-16.lib
+        cp pcre2-32-static.lib pcre2-32.lib
+        cp pcre2-posix-static.lib pcre2-8.lib
```
jelmervdl, line 346: Is this something we should eventually fix upstream? I noticed there [was already something happening to that line](https://github.com/browsermt/ssplit-cpp/pull/10). Since its just bunch of search names, should be fairly easy to search for `-static` or `-staticd` based on `CMAKE_RELEASE_TYPE`.

jelmervdl:
Also once this works (windows builds take ages without cache…) I'd suggest also deleting the `.github/workflows/windows.yml` file. Should be pretty cathartic :)

XapaJIaMnu:
I'll try the official caching. The vcpkg provided on the hosts now is fairly new, it's this commit: https://github.com/microsoft/vcpkg/commit/50fd3d9957195575849a49fa591e645f1d8e7156
I hope it will work with them.

Do you think we should have both avx and avx2 builds? Once upon a time i tested on windows and I didn't see any difference between the two, but maybe that has changed? Now that i have the cmd interface it should be easy enough to test.

The windows build is avx2 always, until we merge the latest bergamot translator (i think).

.github/workflows/build.yml
XapaJIaMnu: I think that Jerin's PR would break compilation even further. A (somewhat improper) fix would be https://github.com/ugermann/ssplit-cpp/pull/29
We shouldn't ever link to a debug PCRE2, if we have a crash in our code, we don't need PCRE2's debug symbols...

jelmervdl:
> Do you think we should have both avx and avx2 builds?

From a user perspective: no. Even I don't have a clue what the machines laying around in this house support. The heavy lifting done by intgemm will detect it at runtime, right? So its only the (hopefully) non-critical paths that would use less optimised instructions if you were to run the avx binary on an avx2-capable machine? I'll definitely boot into Windows on an AVX2 capable machine and see whether I can spot a difference once bergamot-translator/main is merged (working on it!). Until then maybe disable comment out avx builds in the matrix.

.github/workflows/build.yml
jelmervdl: You're right. Probably better to go that route then. I'd assume anyone who has debug libraries installed also has the non-debug versions available. If not, they're probably capable enough to fix that on their own.

XapaJIaMnu:
Hmm, you're correct. I will leave this here until I fix the workflow. Will maybe merge master in here before merging, depending on when i clean up/you clean up the interface.

jelmervdl:
image
😞 No idea why tar failed. Out of disk space maybe?

XapaJIaMnu:
Most likely because of all the extra debug shit compiled... Maybe I should really revisit this thing about having vcpkg.conf

XapaJIaMnu:
Some AVX vs AVX2 results with very rough setup (n = 1)
I run:
`time QT_QPA_PLATFORM=offscreen translateLocally -m en-de-tiny -i 2554-0.txt -o test.avx2`

| TestSet              |   OS   | Host | arch | time  |
|----------------------|:------:|:----:|------|-------|
| Crime and Punishment | ubuntu | zisa | avx  | 43.82 |
| Crime and Punishment | ubuntu | zisa | avx2 | 45.03 |

Time includes model loading time, which is rather slow on the 23-odd cores due to #76 It seems we don't really need avx2 build

XapaJIaMnu:
It's not the "intended" caching way, as I just cache the installed directory, BUT it is proper caching and allows for updates. Merging.

jelmervdl:
🎉🎉🎉

PR #74, opened 2021-12-23
Update bergamot-translator submodule

jelmervdl:
Upgrades the bergamot-translator submodule to include the latest changes.

I know bergamot-translator has a BlockingService, but as far as I understand that one is single threaded. So we're still using the AsyncService in such a way that it is effectively blocking. Until there is an easy way to cancel pending translation requests, this is still the easiest way to get a responsive experience.

### DIFF
--- a/src/MarianInterface.cpp
+++ b/src/MarianInterface.cpp
@@ -4,13 +4,11 @@
 #include "3rd_party/bergamot-translator/src/translator/parser.h"
 #include "3rd_party/bergamot-translator/src/translator/response.h"
 #include "3rd_party/bergamot-translator/3rd_party/marian-dev/src/3rd_party/spdlog/spdlog.h"
-#include "translator/translation_model.h"
 #include <future>
 #include <memory>
 #include <thread>
 #include <chrono>
 #include <QMutexLocker>
-#include <QDebug>
 
 namespace  {
 
--- a/src/MarianInterface.cpp
+++ b/src/MarianInterface.cpp
@@ -4,29 +4,23 @@
 #include "3rd_party/bergamot-translator/src/translator/parser.h"
 #include "3rd_party/bergamot-translator/src/translator/response.h"
 #include "3rd_party/bergamot-translator/3rd_party/marian-dev/src/3rd_party/spdlog/spdlog.h"
+#include "translator/translation_model.h"
+#include <future>
 #include <memory>
 #include <thread>
 #include <chrono>
 #include <QMutexLocker>
+#include <QDebug>
 
 namespace  {
-marian::Ptr<marian::Options> MakeOptions(const std::string &path_to_model_dir, translateLocally::marianSettings& settings) {
-    std::string model_path = path_to_model_dir + "/config.intgemm8bitalpha.yml";
-    std::vector<std::string> args = {"marian-decoder", "-c", model_path,
-                                     "--cpu-threads", std::to_string(settings.cpu_threads),
-                                     "--workspace", std::to_string(settings.workspace),
-                                     "--mini-batch-words", "1000",
-                                     "--alignment", "0.1",
-                                     "--quiet"};
 
-    std::vector<char *> argv;
-    argv.reserve(args.size());
-
-    for (size_t i = 0; i < args.size(); ++i) {
-        argv.push_back(const_cast<char *>(args[i].c_str()));
-    }
-    auto cp = marian::bergamot::createConfigParser();
-    auto options = cp.parseOptions(argv.size(), &argv[0], true);
+std::shared_ptr<marian::Options> makeOptions(const std::string &path_to_model_dir, const translateLocally::marianSettings &settings) {
+    std::shared_ptr<marian::Options> options(marian::bergamot::parseOptionsFromFilePath(path_to_model_dir + "/config.intgemm8bitalpha.yml"));
+    options->set("cpu-threads", settings.cpu_threads,
+                 "workspace", settings.workspace,
+                 "mini-batch-words", 1000,
+                 "alignment", "soft",
+                 "quiet", true);
     return options;
 }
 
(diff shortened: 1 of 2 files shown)


src/MarianInterface.cpp
```diff
@@ -104,34 +99,48 @@ MarianInterface::MarianInterface(QObject *parent)
...
                 } else if (input) {
-                    if (service) {
-                        auto start = std::chrono::steady_clock::now(); // Time the translation
-                        std::future<int> num_words = std::async(countWords, *input); // @TODO we're doing an "unnecessary" string copy here (necessary because we std::move input into service->translate)
+                    if (model) {
+                        std::future<int> wordCount = std::async(countWords, *input); // @TODO we're doing an "unnecessary" string copy here (necessary because we std::move input into service->translate)
 
                         marian::bergamot::ResponseOptions options;
                         options.alignment = true;
-                        options.alignmentThreshold = 0.1f;
```
XapaJIaMnu, line 125: Is this hardcoded now via soft alignment?

src/Translation.cpp
```diff
@@ -107,18 +107,45 @@ QVector<WordAlignment> Translation::alignments(Direction direction, int sourcePo
...
+        
+        // If no alignments were provided by the model, this array will be empty
+        if (response_->alignments[sentenceIdx].empty())
+            continue;
+
+        if (direction == Translation::source_to_translation) {
+            assert(firstWord < response_->source.numWords(sentenceIdx));
+            assert(lastWord <= response_->source.numWords(sentenceIdx));
+
+            for (size_t t = 0; t < response_->target.numWords(sentenceIdx); ++t) {
```
XapaJIaMnu, line 134: Minor: Some partial sorting n-th element could do that, but since it's not too heavy doesn't matter.

src/MarianInterface.cpp
```diff
@@ -4,29 +4,23 @@
 #include "3rd_party/bergamot-translator/src/translator/parser.h"
 #include "3rd_party/bergamot-translator/src/translator/response.h"
 #include "3rd_party/bergamot-translator/3rd_party/marian-dev/src/3rd_party/spdlog/spdlog.h"
+#include "translator/translation_model.h"
+#include <future>
 #include <memory>
 #include <thread>
 #include <chrono>
 #include <QMutexLocker>
+#include <QDebug>
```
XapaJIaMnu, line 13: Extra include forgotten?

src/MarianInterface.cpp
```diff
@@ -4,29 +4,23 @@
 #include "3rd_party/bergamot-translator/src/translator/parser.h"
 #include "3rd_party/bergamot-translator/src/translator/response.h"
 #include "3rd_party/bergamot-translator/3rd_party/marian-dev/src/3rd_party/spdlog/spdlog.h"
+#include "translator/translation_model.h"
+#include <future>
```
XapaJIaMnu, line 8: future is no longer part of the bergamot translator interface so you need to include it?

XapaJIaMnu (approved):
I seem to recall that bergamot translator now supports cache which is off by default. Didn't we need to add some extra options to enable it?

src/MarianInterface.cpp
jelmervdl: Yep, its currently hard-coded in the for-loop, waiting to be replaced by something a bit more dynamic.

src/MarianInterface.cpp
jelmervdl: Oops!

src/MarianInterface.cpp
jelmervdl: I'm using `std::promise` to wait for the `AsyncService::translate()` call to finish. I assumed that none of the other headers would provide it anymore.

jelmervdl:
I haven't tried out the caching thing yet, it's part of AsyncService and can easily be enabled that way. I'll try it out a bit, or we can try it out in a separate thread because it might need a bit more testing? It will also make the benchmarking story a bit more complicated.

I'm recreating the AsyncService (and thus dropping the cache) every time you change the model or any parameters. For changing the model dropping the cache isn't that bad I suppose since you'll be translating to a different language anyway. 

For changing parameters, mainly number of threads, I have to recreate the AsyncService anyway because that's managed in its constructor. Would sort-of be cool if AsyncService had a constructor that would take a cache instance, so we can keep it. Or change the number of workers without recreating it.

jelmervdl:
Cache size should probably be a parameter. I just tried it with War and Peace, and 2k cache entries don't make much of a difference. I've bumped it up to `1 << 16`, which is definitely resulting in a speed-up for follow-up translations. But there are still a lot of cache misses. I'll make a new pull request for it where we can explore it a bit more.

src/MarianInterface.cpp
```diff
@@ -104,34 +97,48 @@ MarianInterface::MarianInterface(QObject *parent)
...
-                    spdlog::drop("general");
-                    spdlog::drop("valid");
+                    service = std::make_unique<marian::bergamot::AsyncService>(serviceConfig);
 
-                    service.reset(new marian::bergamot::Service(MakeOptions(model->config_file, model->settings)));
+                    // Initialise a new model. Old model will be released if
+                    // service is done with it, which it is since all translation
+                    // requests are effectively blocking in this thread.
+                    auto modelConfig = makeOptions(modelChange->config_file, modelChange->settings);
+                    model = std::make_shared<marian::bergamot::TranslationModel>(modelConfig, marian::bergamot::MemoryBundle{}, modelChange->settings.cpu_threads);
```
jerinphilip, line 115: The WASM optimized load if I remember correctly is via `MemoryBundle` which appears to be empty here and FS loads are happening. FS loads have been slow (not just WASM). 

@XapaJIaMnu #76.

jerinphilip: https://github.com/jerinphilip/lemonade/blob/b8f4264084127e3da2b07ef280de8fc9fc5db725/lemonade/lib/translator.cpp#L28-L36, although I'm still not particularly happy with the load times after either. But should be faster than repeated on-disk loads.

jelmervdl: Thanks! To me the difference seems to be reading once, and reading once for every worker thread. I'll implement it that way, see how well it works. I assume it just reads the file into memory, and each worker still has to do the parsing on their own, but changing that (having all workers share those resources?) seems … nontrivial.

PR #75, opened 2021-12-23
Enable caching

jelmervdl:
bergamot-translator has a cache implementation. This would be really helpful for interactive translations where most of the sentences won't be affected by a local edit.

- [x] Enable cache
- [x] Pick a default cache size [1]
- [x] Make cache configurable [2]

[1] The cache implementation does not do any probing, so if there is a hash collision you get only one of them entries making it into the cache. So a small size means many collisions, since `index = hash(sentence) % size`.

[2]  I'm thinking a drop-down in the prefs that goes from 0 (disabled) to something big, but have options like "2mb" "200mb" etc. Number of entries doesn't tell you much anyway since the implementation does not guarantee that those entries will be used.

### DIFF
--- a/src/MarianInterface.cpp
+++ b/src/MarianInterface.cpp
@@ -141,11 +141,6 @@ MarianInterface::MarianInterface(QObject *parent)
                         double words = wordCount.get();
                         std::chrono::duration<double> elapsedSeconds = end-start;
                         int translationSpeed = std::ceil(words/elapsedSeconds.count()); // @TODO this could probably be done in the service in the future
-
-                        qDebug() << "Cache stats:"
-                                 << "\n  hits =" << service->cacheStats().hits
-                                 << "\n  misses =" << service->cacheStats().misses;
-                        
                         emit translationReady(Translation(std::move(future.get()), translationSpeed));
                     } else {
                         // TODO: What? Raise error? Set model_ to ""?
--- a/src/MarianInterface.cpp
+++ b/src/MarianInterface.cpp
@@ -9,6 +9,7 @@
 #include <thread>
 #include <chrono>
 #include <QMutexLocker>
+#include <QDebug>
 
 namespace  {
 
@@ -102,6 +103,9 @@ MarianInterface::MarianInterface(QObject *parent)
                     // @TODO: don't recreate Service if cpu_threads didn't change?
                     marian::bergamot::AsyncService::Config serviceConfig;
                     serviceConfig.numWorkers = modelChange->settings.cpu_threads;
+                    serviceConfig.cacheEnabled = true;
+                    serviceConfig.cacheSize = 1 << 24;
+                    serviceConfig.cacheMutexBuckets = modelChange->settings.cpu_threads;
                     
                     // Free up old service first (see https://github.com/browsermt/bergamot-translator/issues/290)
                     service.reset();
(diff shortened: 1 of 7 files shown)


XapaJIaMnu:
yeah, the whole segmentation bit makes it difficult. I vaguely recall @kpu  complaining that `std::hash` doesn't do a good job at distributing values evenly across a uint64_t. This is not something we can solve here though.

jelmervdl:
In any case, with the high cache limit I've configured now its no longer translation that makes editing War & Peace unpleasant… It's the Qt part that makes the interface slow down. The experience is almost a step backwards because now typing is laggy. So that's a win, I guess.

bergamot-translator uses [the vocab positions](https://github.com/browsermt/bergamot-translator/blob/9e1c1e8dbf4817f411f718c75ce42493d92c43b6/src/translator/request.cpp#L15) combined through a function [made for combining larger hashes](https://github.com/browsermt/marian-dev/blob/08b1544636fe13eaf1fbacb17c6fb050abfb8d42/src/common/hash.h#L14) so maybe there is room for some easy improvements.

Is there an order to the vocabulary? I assume it is somewhat based on frequency. If that's the case, maybe just `std::hash(index)` would already be an improvement. If the order is random, some cheap operation to map 0..vocabsize to 0..numeric_limits<size_t>::max() could work?

XapaJIaMnu:
Yes, the vocabulary in 99% of the cases (Unless someone specifically went out of their way to create custom vocab) is ordered based on word frequency with the 0-th and 1-st items reserved.

I think users dislike laggy typing more than waiting for translations to happen ;/

jelmervdl:
For reference, cache usage when translating and re-translating Crime & Punishment:
image

I can't see a measurable difference in memory usage in Activity Monitor, no matter which of the cache sizes I try.

XapaJIaMnu:
Did you try varying the bucket size?

jelmervdl:
I haven't. I've just set mutexBuckets_ to the number of threads. That might impact speed a bit (because lock contention) but I doubt it will be much.

If you ment buckets behind the index to catch those initial hash-based collisions, [the cache implementation](https://github.com/browsermt/bergamot-translator/blob/main/src/translator/cache.h) is simple, there are no buckets behind the initial index.

XapaJIaMnu:
Oh yeah, I remember now that the buckets didn't do that much...

We can try to go with murmur hash, but even without we seem to be getting decent scores

jelmervdl:
I was working on making the cache size configurable, but having a hard time with the labels (like: how do you communicate "this setting will allow you to cache about 1.000.000 sentences in theory but more like 10.000 in practise and will cost you about 5mb in theory, but it depends on the average length of your sentences and whether you're typing the whole thing or just pasting it as is. Also the 5mb will not be noticeable because other parts of our app use much larger dynamic allocations. Also behaviour will be different when you use translate-as-you-type as you'll fill the cache with useless translations quicker. Also cache won't help when you type only a single sentence. What is a sentence you ask?"

In other words:
- how much would such a setting add? 
- can we get away with just picking a good default?
- Is there a scenario in which we would want to disable the cache?

XapaJIaMnu:
I don't think this parameter should be user configurable. Modern systems have at the very least 4 GB of Ram. Those machines would likely not have more than 4 threads, which means we'd default to 2 workers with about 300 MB of memory usage. 5-10-15 megabytes would not break the bank. Maybe just have disable cache with a note (Disable cache if you experience lag in GUI) ;D?

XapaJIaMnu:
So i tried, it. On my 7700HQ I don't have any GUI lag when post-editing Crime and punishment (Although the fans do go crazy).
The only issue is that when I post edit, when the translation is updated, the text on the target side jumps ahead and i can no longer see the sentence I just post-edited. Clicking anywhere in the text immediately fixes the alignment and correctly displays the target portion.

Is this an issue of how we approximately calculate the scrolling?

jelmervdl:
> Is this an issue of how we approximately calculate the scrolling?

I think so. Right now we just interpolate the target scroll position, i.e. `output-scroll-pos = input-scroll-pos / input-vertical-length * output-vertical-length`. But that might go a bit wonky when you're editing at the bottom of the text field (which is most often the case…). I'd suggest making another issue for that, I don't think it's related to caching. Maybe just more noticeable because translation is faster, so you're less ahead of the translation when the scroll position is queried & copied.

PR #80, opened 2022-01-04
Interruptible translations

### LINKED ISSUE #78: force-stop translation on quit
I just noticed that ctrl+c on the command-line does kill in-flight translation, but Cmd+Q keeps the application alive till translation is finished. It does close the main window though.

To reproduce: select a model, paste in Crime & Punishment, press cmd+q before it finishes translating.

Result: The main window closes immediately, but the application stays active in the foreground and in the dock.

Expected: application closes completely when the main window disappears.

Not sure if there is a way to somehow either signal to bergamot-translator that it is time to stop, or signal the MarianInterface worker that it (and all its "child threads") should just bail. Since they're not separate processes I don't think there is an easy way to achieve this?

jelmervdl:
Fixes #78.

This implements a way to interrupt in-progress translations to make switching models and quitting TranslateLocally much faster by not having it wait for a (long!) translation to finish.

Unfortunately I needed to "roll my own promise" because I couldn't figure out an easier way to wait on a future while also being interruptible in case the future was cancelled. Qt's futures can be cancelled but have the annoying bit where you have to call `finish()` on them for the blocking call `waitForFinished()` to return, even when cancelled. But you can't copy them, so you end up having to put them in a `shared_ptr`. In other words: more complex than this simple struct.

Depends on https://github.com/browsermt/bergamot-translator/pull/297


src/Promise.h
```diff
@@ -0,0 +1,36 @@
...
+#include <QSemaphore>
+
+// Simple cancelable(!) promise because the machinery to do this
+// using std::future is still experimental. Also, QFuture would be
+// an option, except that it requires you to call finish() after
+// cancelling (and explicitly forbids finishing multiple times)
+// which makes using it in an not-necessarily-called callback
+// hard.
+template<typename T>
+struct Promise {
```
XapaJIaMnu, line 12: I don't like the promise name here because, https://en.cppreference.com/w/cpp/thread/promise

I mean, I know it's a promise, but when I was reading the above code I thought we were using std::promise. Maybe Just rename it as OurPromise or CustomPromise?

src/MarianInterface.cpp
```diff
@@ -119,26 +119,37 @@ MarianInterface::MarianInterface(QObject *parent)
...
-                        std::promise<marian::bergamot::Response> response;
-                        auto future = response.get_future();
+
+                        // Prepare our translation promise. We use this to wait on
+                        // either the translation to finish, or be cancelled. Hence
+                        // using member variable pendingTranslation_ to make it
+                        // accessible from outside the worker.
+                        {
+                            QMutexLocker locker(&lock_); // brief lock because changing member variable
+                            pendingTranslation_ = std::make_shared<Promise<Translation>>();
```
XapaJIaMnu, line 129: Is this because any worker could take that translation?

XapaJIaMnu (commented):
The complexity of the main working function is really high with additional mutexes. Do you think we could simplify it?

src/Promise.h
jelmervdl: I'll rename it to CancelablePromise. That's its raison d'être.

src/MarianInterface.cpp
XapaJIaMnu: no, the workers should inside the translation service, I don't quite understand this?

XapaJIaMnu: Async translation submissions from the GUI?

jelmervdl: MarianInterface has a single worker internally, but that one will block during translation. So the CancelablePromise is exposed to the part of MarianInterface that talks to the main thread (the Qt GUI thread) so that it can interrupt the worker.

I'm briefly locking here to have exclusive access to `pendingTranslation_`. It's a bit vague what happens when you'd trigger the old's promise cancel() just before this section. Then that interruption would be ignored I suppose…

src/Promise.h
XapaJIaMnu: BreakablePromise? Too dramatic?

src/MarianInterface.cpp
jelmervdl: > Async translation submissions from the GUI?

I want that, but I don't want the backlog. Say you're typing and you just keep submitting stuff:
```
I                    (requested)
I want               (requested)
I want cake          (requested)
I want cake and      (requested)
                     (translation of "I" appears)
I want cake and eat  (requested)
                     (translation of "I want" appears)
```

With the current construction you have:
```
I                    (requested)
I want               (replaced)
I want cake          (replaced)
I want cake and      (requested)
                     (translation of "I" appears)
I want cake and eat  (requested)
                     (translation of "I want cake and" appears)
```

It's a bit like double buffering and skipping frames instead of slowing the whole game down because you can't match the 30hz refresh rate.

You might be able to get to this point with a light promise wrapper around bergamot-translator (to turn the callback into promises, sorry Jerin!) but you'd still have a bit of a mess around the logic of when to load a model (off the main thread otherwise that blocks for a second!), when to submit a translation, when to queue but not submit a translation, etc. MarianInterface effectively is there now to hide that mess from the rest of TranslateLocally.

src/Promise.h
jelmervdl:
```cpp
template<typename T> struct WeddingVow {};
```
You can break it, but AsyncService will leave you and you can never come back with new translation requests.

XapaJIaMnu: I like that.

src/MarianInterface.cpp
XapaJIaMnu: Yeah, we should only ever submit one thing to the translator, as we do it now. I just wonder if we can somehow simplify the logic because that while loop is on the way to being unmaintainable.... BUt let's add the functionality we want first and then we can take a look at how the logic can be simplified.

jelmervdl: I've reworked it a bit to use condition_variable instead. No need for a promise this way, and it is much more clear what conditions would cause it to continue the loop.

The loop is still broken into a "swap pointers with internal pointers so I can release the lock" and "do actual work" section. If you know a good way to simplify that part, I'm very interested!

jelmervdl:
If this were to be merged, please squash it to hide my indecisiveness plz 😅

XapaJIaMnu:
I read through the code and I understand what's happening. Do you want to merge?

jelmervdl:
I'd say it should wait till https://github.com/browsermt/bergamot-translator/pull/297 is approved & merged. This doesn't add much without it.

Other than that I'm not 100% sure I used std::condition_variable in a legal way (the one in the second half of the while loop is linked to a different mutex). If someone who is more familiar with them could approve what I'm doing to tell me that this is stupid, I'd really appreciate that.

XapaJIaMnu:
Afaik you can? https://stackoverflow.com/questions/57842147/can-you-signal-a-condition-variable-from-a-different-mutex-from-the-one-that-is

XapaJIaMnu:
I think this can be merged. The only thing is, it took me a while to understand that `emit pendingChanged(true)` controls the working bar in the GUI...

PR #81, opened 2022-01-18
Rich text formatting

jelmervdl:
First steps towards working with text with markup info. End goal is to be able to copy/paste a (part of a) document from Microsoft Word into TranslateLocally, have it translated, and be able to copy/paste the output directly back into Word and be it with the right formatting.

- [x] Maintain markup through translation
  - [x] Expand bergamot-translator to accept xhtml-style self-closing tags :(
- [ ] Toggle between rich text and plain text
- [ ] Have controls/keyboard shortcuts to edit markup (e.g. ctrl + b to make something bold)
- [ ] (Is it just me or is it slow?)

### DIFF
--- a/src/MarianInterface.cpp
+++ b/src/MarianInterface.cpp
@@ -117,9 +117,14 @@ MarianInterface::MarianInterface(QObject *parent)
                     model = std::make_shared<marian::bergamot::TranslationModel>(modelConfig, std::move(bundle), modelChange->settings.cpu_threads);
                 } else if (input) {
                     if (model) {
+                        // Remove the "<!DOCTYPE html>" bit
+                        auto begin = input->find("<html");
+                        input->erase(0, begin);
+
                         std::future<int> wordCount = std::async(countWords, *input); // @TODO we're doing an "unnecessary" string copy here (necessary because we std::move input into service->translate)
 
                         marian::bergamot::ResponseOptions options;
+                        options.HTML = true;
                         options.alignment = true;
                         
                         // Using promise for a translation, and future for waiting
--- a/src/Settings.cpp
+++ b/src/Settings.cpp
@@ -16,7 +16,8 @@ Settings::Settings(QObject *parent)
 , showAlignment(backing_, "show_alignment", false)
 , alignmentColor(backing_, "alignment_color", QColor("#EDD400"))
 , syncScrolling(backing_, "sync_scrolling", true)
-, windowGeometry(backing_, "window_geometry") {
+, windowGeometry(backing_, "window_geometry")
+, textFormat(backing_, "text_format", AutoDetect) {
     //
 }
 
--- a/src/Settings.h
+++ b/src/Settings.h
@@ -93,4 +93,12 @@ public:
     SettingImpl<QColor> alignmentColor;
     SettingImpl<bool> syncScrolling;
     SettingImpl<QByteArray> windowGeometry;
+
+    enum TextFormat {
+        PlainText,
+        RichText,
+        AutoDetect // plainText unless it switches to RichText
+    };
+
+    SettingImpl<TextFormat> textFormat;
 };
--- a/src/mainwindow.cpp
+++ b/src/mainwindow.cpp
@@ -121,7 +121,7 @@ MainWindow::MainWindow(QWidget *parent)
             // can't reach it with cursor keys, but it does show up when you use
             // the scrollbar. So to match the line count better, also add it to
             // the output.
-            ui_->outputBox->setPlainText(translation_.translation() + QString("\n"));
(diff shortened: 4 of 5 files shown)


src/MarianInterface.cpp
```diff
@@ -117,9 +117,14 @@ MarianInterface::MarianInterface(QObject *parent)
                     model = std::make_shared<marian::bergamot::TranslationModel>(modelConfig, std::move(bundle), modelChange->settings.cpu_threads);
                 } else if (input) {
                     if (model) {
+                        // Remove the "<!DOCTYPE html>" bit
+                        auto begin = input->find("<html");
+                        input->erase(0, begin);
```
jerinphilip, line 122: Any reason why [we](https://github.com/jerinphilip/tagtransfer/blob/558ff384ef8452b3abb7a64b3dd917149f7a0abf/tagtransfer/html_translator.py#L21-L29) shouldn't be doing this at bergamot-translator? This is nice to have for everyone trying to translate HTML with our library, is it not?

XapaJIaMnu:
Hmm, I can't run it.  I get stuck at this screen on Startup:
Screenshot_20220120_130542

QT5 issue?

jelmervdl:
> Hmm, I can't run it. I get stuck at this screen on Startup: Screenshot_20220120_130542
> 
> QT5 issue?

Looks like you're missing the bergamot-translator update that makes it accept self-closing tags. Did you update the submodule to the specific checkout (or main, since it has been merged already)?

XapaJIaMnu:
You were right, i didn't run `git submodule update --recursive`
BBC looks... Pretty bad:
Screenshot_20220120_131421

Wikipedia gives me an error:

Screenshot_20220120_131523

Also, Maybe in this mode, we should paste images for it to look better?

https://stackoverflow.com/questions/3254652/several-ways-of-placing-an-image-in-a-qtextedit

But in general web pages look horrible even when pasted in more complicated editors sometimes:

Screenshot_20220120_133114

jelmervdl:
The HTML supported by Qt's TextEdit is quite limited. It basically is some inline stuff that is translatable to their internal markup. Nothing more. So copy/pasting most pages with CSS will not do anything good.

Looking at that error though, I'm starting to see trouble. It might be pretty difficult to implement this properly using the HTML support we have in bergamot-translator. Looking at [how Qt turns markup into HTML](https://github.com/qt/qtbase/blob/9a9b253b686618e23c2874cbcb2e5b89e56cd66d/src/gui/text/qtextdocument.cpp#L2969) it looks like they don't emit closing tags for many elements, such as `<li>`. If we want this, we might need to write our own QTextDocument -> HTML -> QTextDocument serialisation pipeline.

(Also looking at that code, I think things break down if you put HTML into the document title because [that is added unescaped?](https://github.com/qt/qtbase/blob/9a9b253b686618e23c2874cbcb2e5b89e56cd66d/src/gui/text/qtextdocument.cpp#L2369-L2371))

XapaJIaMnu:
Hmm... In that case we should probably disable handling html and just handle xml from like word documents? Can we detect that by peeking in the clipboard.

jelmervdl:
> Hmm... In that case we should probably disable handling html and just handle xml from like word documents? Can we detect that by peeking in the clipboard.

It's not our support for HTML, it's Qt's support for markup. When you copy/paste something into TranslateLocally, it is converted into Qt's own markup structure (QTextDocument, which is a bunch of QTextBlocks). What they support with that markup is limited. Whether you copy in from a Word document or a website, it will always be limited to what that data structure can represent. Any other markup is lost.

The error message you ran into was due to how Qt does QTextDocument -> html:str. The screenshots though, those are all just illustrations of the limitations of QTextDocument.

src/MarianInterface.cpp
jelmervdl: You're right, we probably should. [It's not difficult to parse](https://html.spec.whatwg.org/multipage/syntax.html#the-doctype). But then we also have to support adding it back into the translated output, right?

XapaJIaMnu:
Hmmmm..... What if we use QtWebKit for rendering? That at the very least would have a much better HTML support, but looking into it, it will not be a trivial task. I am also worried about overengineering though. 

At the moment it seems like enabling rich text will work in a very few limited cases. My idea about peeking was mostly to see if we have XML content that we can handle before enabling RTF mode, else fall back to plain text mode?

Godnoken:
I'm sorry if this is not on topic as you are talking about HTML primarily.

I noticed that I can't successfully translate texts that contain \n through the command line
I "fixed" it by substituting \n with another symbol like * during translation and then vice versa after the translation is done. It works sometimes but often not, due to my text input having different sets of \n, \r and so on.

Is there or would there be a way to ignore all sorts of carriage returns?

PR #92, opened 2022-02-03
Better CLIOnly support in display server-less environments

XapaJIaMnu:
We create a QCoreApplication to do the command line parsing and then run a CLI only application if we are supposed to run a CLIOnly application, else we run a GUI application.

@jelmervdl  how about this. The only thing I am not sure about is skipping the CLI parsing in GUI mode. @jerinphilip that should resolve the issue of needing to specify a QT plugin to run in terminal.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -264,7 +264,7 @@ jobs:
     - name: Output test # @TODO some more complicated test, perhaps a unit test
       working-directory: ${{github.workspace}}/build
       shell: bash
-      run: ./translateLocally --version # Can't run normally due to not having X server set up
+      run: ./translateLocally --version
 
     - name: Package # Produce a .deb file using cpack
       working-directory: ${{github.workspace}}/build
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -196,7 +196,7 @@ jobs:
     - name: Dependencies  # @TODO should also test qt6, but it's not present yet
       run: |
        sudo apt-get update
-       sudo apt-get install -y xvfb libpcre++-dev qttools5-dev qtbase5-dev libqt5svg5-dev libarchive-dev libpcre2-dev ccache
+       sudo apt-get install -y libpcre++-dev qttools5-dev qtbase5-dev libqt5svg5-dev libarchive-dev libpcre2-dev ccache
       
       # https://software.intel.com/content/www/us/en/develop/articles/installing-intel-free-libs-and-python-apt-repo.html
     - name: Install MKL
@@ -264,7 +264,7 @@ jobs:
     - name: Output test # @TODO some more complicated test, perhaps a unit test
       working-directory: ${{github.workspace}}/build
       shell: bash
-      run: xvfb-run --auto-servernum ./translateLocally --version # Can't run normally due to not having X server set up
+      run: ./translateLocally --version # Can't run normally due to not having X server set up
 
     - name: Package # Produce a .deb file using cpack
       working-directory: ${{github.workspace}}/build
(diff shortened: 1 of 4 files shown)


jelmervdl (approved):
Works like a charm! I tested both the signed and unsigned macOS bundles.

I assume you've tested it in an X-less environment. I was thinking of building it inside a docker container but effort if you've already tested it.

jelmervdl:
Oh one test you could do is update the Github actions in this branch to not use the fake-X-server. That would verify what I tried to do _and_ set up a good automated test for it :)

XapaJIaMnu:
Yeah I tested it already on my machine, but you're right I need to update the workflow. Also we should have a proper test now that CLI mode is working well. I will write something of the sort "download a model, translate one sentence, then die"

jerinphilip:
> The only thing I am not sure about is skipping the CLI parsing in GUI mode

Does this remove any possibilities of launching translateLocally GUI via pipe and a few args? Is it even a thing anymore to such? Right-click to edit in translateLocally or something like iBus native spawning translateLocally for a verification UI if we ever get to that?

XapaJIaMnu:
At the moment we only use the CLI for CLI translation. There's no way to launch a GUI and also perform a CLI operation in the background, so that's why i didn't do CMD parsing in the app as it's not necessary. We might want to add some switches in future versions, but I guess when it comes to that we will re-enable CLI parsing in the main? Or should I just re-do CLI parsing just in case something happens. I mean, it's not expensive or anything.

jelmervdl:
It would be pretty easy to implement some flags that make `isCliOnly()` return false, and then have GUI-specific cli parser that interacts better with the GUI in a future feature.

I quite like how we can switch modes this way. Adding a nativemessaging server behind a command line flag that would leverage bergamot's AsyncService will be pretty easy. Really because we can have different modes with basically their own main() function.

XapaJIaMnu:
I agree with @jelmervdl , we can make a future non-CLI parser if we so desire. I will merge this once windows builds succeed.

jerinphilip:
While we're waiting for Windows builds, any angles of getting rid of vcpkg? Everything else appears to be succeeding fast.

XapaJIaMnu:
vcpkg is the only way to get a static qt build on windows. Sorry ;/ My caching used to work, but it seems to break every once in a while and i have no idea why...

jelmervdl:
We use actions/cache@v2, the documentation suggests using Github's Packages thingy: https://github.com/microsoft/vcpkg/blob/master/docs/users/binarycaching.md#binary-caching

But I know how much time @XapaJIaMnu spent on getting the windows builds to work at all… I wouldn't suggest anyone to play with that unless they really, really have nothing better to do and are mentally prepared for it.

XapaJIaMnu:
The example in the documentation, probably works, but I didn't understand how to do it outside of nuget, and nuget is paid.
Technically it should be possible but I need to boot up a windows machine and spend a day trialing and erroring vcpkg to get it right... And I have a life too you know.

jerinphilip:
Most of my ask is for ideas or information you already know and can speed up a follow-up attempt, not immediate demands to solve. 

> I wouldn't suggest anyone to play with that unless they really, really have nothing better to do and are mentally prepared for it.
> Technically it should be possible but I need to boot up a windows machine and spend a day trialing and erroring vcpkg to get it right...

I understand we're positioning translateLocally to be a part of native messaging solution. vcpkg is also a problem at bergamot-translator (+5m so milder than +2hour turnaround here). At least I like fast turnarounds and have a windows machine, so some curiosity to poke and motivation to take this on at some point. Let's hope that happens. xD

kpu:
The punishment for raising unrelated issues like vcpkg is that the issue gets closed when the original problem is solved.

PR #93, opened 2022-02-11
Initial native message iface

XapaJIaMnu:
I will develop a more complete native message interface, with json communication that will accept commands from the browser. For now though it is a simple stdin listener...

To test you can use the following python code:
```
import ctypes
def encode_msg(message: str) -> bytearray:
	msglen = len(message.encode('utf-8'))
	len_in_bytes = bytes(ctypes.c_int(msglen))
	msg_in_bytes = bytes(message.encode('utf-8'))
	return len_in_bytes + msg_in_bytes

In [1]: encode_msg("Hello, world")
Out[1]: b'\r\x00\x00\x00Hello, world'
```
Then you get to experience some buffer overflow because it's 1:30 AM and I will fix it tmrw..

```bash
% echo "\x0c\x00\x00\x00Hello, world" | ./translateLocally -p                                                                                                                                              :(
Loading model: English-German tiny
Native msg interface accepting messages...
Received message size of: 12
Received message size of: 12 with content: Hello, world
Hallo, Welt
Received message size of: 10
Received message size of: 10 with content: ���

Received message size of: 10
Received message size of: 10 with content: ����
```

Needless to say this is an early draft and has a lot of space for improvements.

I propose that this client will accept the following json encoded commands:
1. What models do you have?
2. What remote models do you have?
3. Load  model X.
4. Download and load model Y.
5. Translate.
6. Die.

It will also respond to all of them with some well formatted json that we can read in the browser.
@kpu @jelmervdl @jerinphilip comment on json.


src/cli/NativeMsgIface.cpp
```diff
@@ -0,0 +1,59 @@
...
+        std::cin.read(len, 4);
+        unsigned int ilen = *reinterpret_cast<unsigned int *>(len);
+        if (ilen < kMaxInputLength && ilen>1) {
+            std::cerr << "Received message size of: " << ilen << std::endl;
+            // This will be a json which is parsed and decoded, but for now just translate
+            std::unique_ptr<char[]> input(new char[ilen]);
+            std::cin.read(input.get(), ilen);
+            std::cerr << "Received message size of: " << ilen << " with content: " << input.get() << std::endl;
+            translator_->translate(QString(input.get()));
+            eventLoop_.exec(); // Do not read any input before this translation is done
```
jerinphilip, line 35: So we're doing serial? Perhaps allow multiple inputs and opts in this case then?

jelmervdl:
In my extension I skipped the "load model" and "download & load model" messages because that's state that would need to be synced between both sides that are talking. However, it is useful to have these to "pre-load" the model before you start sending translation input. So let's keep them, but make sure that multiple calls to them, i.e. when the model is already loaded, don't cause trouble. Same for calling a _download & load model_ when that specific model (and version) is already downloaded.

Some message for getting a list of available models is great. If that doesn't scale (because too many models when you include remote repositories maybe?) you could limit it by search for source and target languages (lists for each sides though, because the extension often has a list of languages it thinks the page can be in, and same for languages the user might like.) 

I don't think there is need for a `die` message. I'd say closing stdin is enough of a signal.

Let's not use `MarianInterface` here. That one is specifically optimised for latency, and cancels all intermediate translation requests (and cannot batch multiple translation requests together) which is not ideal for a plugin-type usage 

I would also want to propose to make all messages async. I.e. each message received by this endpoint has a message id, and the response will also have that same message id, but there is no guarantee that the order of requests and responses are the same. This way we can
1. much better make use of `AsyncService` etc, 
2. do better batching on the C++ side instead of the javascript side, 
3. have translation responses returned as fast as they're available (instead of having a `BlockingService` like API), 
4. implement translation request cancellation at some point.

src/cli/NativeMsgIface.cpp
XapaJIaMnu: I was imagining js doing the batching, but as you and Jelmer pointed out, probably it's best to do it on the CPP side. Will change.

kpu:
Yes, async is better.  

While there should be a managed default per language pair, language codes shouldn't be the sole identifier of a model.  We already have speed variants and there will soon be OPUS overlap too.  Search by pair makes sense here: that way pivoting and multilingual models can be hidden from the client rather than enumerating a long list of possibilities.  

A load message is good to reduce latency when typing happens.  Recommend combining the "load model" and "download & load model" message so it's identical from the client.  And the message should be idempotent as Jelmer suggests.  Response could be download progress, loaded, or error.

XapaJIaMnu:
Suggestions heard, stand by later today for async variants. @jelmervdl I will keep using marianIface for now and replace it at a later point.

XapaJIaMnu:
First MVP-ablae version. It reads and writes json, including e mssage id (although I think I don't set the id yet, but that's easy to fix).

What is missing: 
1) Termination condition. Is there a way of knowing when `service` has finished all of its work and doesn't have any pending threads? I remember @jelmervdl had similar concerns when implementing quit immediately?  At the moment, if a termination signal is sent, all threads will be killed immediately and any in flight translations will never  arrive.

2) Constructing models. Currently everything is hardcoded. Need to implement a model change, which is all conditioned on 1).

3) What json do we actually send back. Response is a pretty big object, do we need all of its entries?

jelmervdl:
> Is there a way of knowing when service has finished all of its work and doesn't have any pending threads?

AsyncService's destructor will block until all translation requests are processed and all workers have finished their tasks. So you can just let it go out of scope (or reset a unique ptr or something, anything that causes its destructor to be called) and after that you're safe.

> Constructing models. Currently everything is hardcoded. Need to implement a model change, which is all conditioned on 1).

With AsyncService you don't need to wait for the service to finish. You can just load up another model, and when queueing new translation requests use that one. If you want to release models that are no longer used you can release its shared pointer (all pending requests that still need it will have a shared_ptr to it, so it will be released once that's safe to do.) (but how do you know that the next request won't need them)

> What json do we actually send back. Response is a pretty big object, do we need all of its entries?

For practical purposes, I'd only need `{id: Number, target: String}` at the moment. But if you want to use alignment info, you'd also want `source` and `alignments` obviously. But if you expose id + target, I can try to plug the thing into the firefox extension as an experiment.

jerinphilip:
I have the same inputs as @jelmervdl. However, I want to do something "incremental" for the following:

>> What json do we actually send back. Response is a pretty big object, do we need all of its entries?
> For practical purposes, I'd only need {id: Number, target: String} at the moment.

I gave this some thought while initially referring to https://github.com/mozilla/bergamot-translator/wiki/Unified-API-Example designing `Response`, keeping in mind a JSON export and preparing https://github.com/jerinphilip/lemonade/blob/7eca7e425b027b445a0047b1191d2989f9b0494c/lemonade/lib/json_interop.h#L45-L87. I propose to go with the following in the first iteration for `toJSON(Response)`:

```json
{
	"source": {"text": "..."},
	"target": {"text": "..."}
}
```

We may drop the "source" key in the first iteration but the structure needs to incrementally grow and mimic C++ Response, while consistent across JSON, C++, WebAssembly and Python bindings. 

I have also opened browsermt/bergamot-translator#350 to make WebAssembly consistent with the JSON so both can operate easily.

XapaJIaMnu:
@jerinphilip , we had a discussion offline with @jelmervdl  about the stdout buffer size: https://stackoverflow.com/questions/10904067/in-c-whats-the-size-of-stdout-buffer
It's much smaller than what I thought and it is also difficult to judge what the size is going to be on different platforms and different browsers. Currently, each worker executes a callback that writes to stdout before pushing for a new translation. A bad client that sends many small requests but doesn't read stdout before all data is sent will cause our translation threads to stall until the buffer is cleared.. 

This is why I am questioning what strictly needs to be in the json. It is otherwise relatively trivial for me to put extra things in there.

jelmervdl:
Oh but it is more fun than that! I just noticed:

> The maximum size of a single message from the application is 1 MB. The maximum size of a message sent to the application is 4 GB.

Source: https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/Native_messaging#app_side

At least Firefox seems to be doing concurrent reading & writing using async IO so no deadlock because "it can't read from stdin while it is still trying to write to stdout" on our side I expect. (I haven't looked at other browsers.)

```js
this.readPromise = this.proc.stdout
      .readUint32()
      .then(len => {
        if (len > NativeApp.maxRead) {
          throw new ExtensionError(
            `Native application tried to send a message of ${len} bytes, which exceeds the limit of ${NativeApp.maxRead} bytes.`
          );
        }
        return this.proc.stdout.readJSON(len);
      })
      .then(…)
```

Source: https://searchfox.org/mozilla-central/source/toolkit/components/extensions/NativeMessaging.jsm#165.

So it seems long term we might need to add some sort of multipart messaging response messaging which I'm really dreading. How are you going to break up a large alignments table that in itself exceeds 1mb when encoded as JSON? Because the messages till have to each be valid JSON 😭 

For an MVP I doubt that a Firefox extension will hit that limit, it wants small chunks of translated text anyway to make it feel more like loading a progressive JPEG.

jerinphilip:
> How are you going to break up a large alignments table that in itself exceeds 1mb when encoded as JSON?

To be clear, I am ideally not looking at exporting the hard-alignments table. But life is simpler if I have a JSON representation of the struct so I quickly know what to edit. Alignments are in response consumption at HTML, we're just choosing not to clear it out because WebAssembly VM would have already allocated it. 

The simpler version of what I'm trying to say is keep the keys you want from the existing composition of `Response` structure. Leave the remaining out for later addition if possible. 

`target: String` vs `{"target": {"text": "..."}}` is not going to make much of a difference, but the consistency will help me easily develop at least.

XapaJIaMnu:
If we want the NativeInterface response json to 100% resemble response itself, the proper thing to do would be for `toJson()` method to be added to the Response. That way a change in the API would not require code changes in multiple places.

jelmervdl:
> If we want the NativeInterface response json to 100% resemble response itself, the proper thing to do would be for toJson() method to be added to the Response. That way a change in the API would not require code changes in multiple places.

True, except that ideally we don't want to deal with JSON strings in most interface implementations. The Python bindings convert a response to native python objects. WASM hides the serialisation/deserialisation in the emscripten bindings (which does a lot on demand instead of up front). The part where we communicate with a JSON string is specific to this interface. Just the names of things, and how they are nested, is the same across all interfaces.

XapaJIaMnu:
@jelmervdl  I see. The response struct has multiple sub-structs and some of them have sub-structs. Is there one place where I can view the whole scheme of the configuration?

jelmervdl:
I don't have one 😅 Maybe @jerinphilip does. Things like the Annotation part of an AnnotatedText I have never used outside C++, so I don't know what the interface to that would be. Internally it works with a sort of byte offset array, but I'm not sure we would want to expose that implementation detail via JSON… I suppose figuring this out is worth an issue on its own.

For now, really the only important bit is `{"target": {"text": String}}`. That matches what is in Response, and is the only thing necessary to test implementations.

XapaJIaMnu:
@jelmervdl currently it looks like:
```bash
./native_client.py                                                                                                                                                               
(None, None)
b'/\x00\x00\x00{"text": "Hello world!", "id": 0, "die": false}e\x00\x00\x00{"text": "Sticks and stones may break my bones but words WILL NEVER HURT ME!", "id": 1, "die": false}?\x00\x00\x00{"text": "Why does it not work well...", "id": 2, "die": false}'
Loading model: English-German tiny
{
"id": 2,
"target": {
"text": "Warum funktioniert es nicht gut..."
}
}

{
"id": 0,
"target": {
"text": "Hallo Welt!"
}
}

{
"id": 1,
"target": {
"text": "Stöcke und Steine können mir die Knochen brechen, aber Worte werden NIEMALS HURT ME!"
}
}
```
Tomorrow, I will add model loading functionality.

jelmervdl:
Few tweaks, but I'm getting somewhere with this!
image

(Don't worry about that `Ã` situation, that might be because I'm giving it HTML to translate without `ResponseOptions.HTML=true`)

jelmervdl:
I've pushed some additions that were necessary to get it to work with Firefox, but this now works with https://github.com/jelmervdl/firefox-translations/tree/native-messaging (given you go to a page that matches the model that translateLocally selected)

XapaJIaMnu:
@jelmervdl @jerinphilip model  loading support and pivoting, error reporting.
```bash
% ./native_client.py
b'I\x00\x00\x00{"text": "Hello world!", "id": 0, "die": false, "src": "en", "trg": "de"}\x7f\x00\x00\x00{"text": "Sticks and stones may break my bones but words WILL NEVER HURT ME!", "id": 1, "die": false, "src": "en", "trg": "es"}a\x00\x00\x00{"text": "\\u00bfPor qu\\u00e9 no funciona bien?", "id": 2, "die": false, "src": "es", "trg": "de"}'
Received message size of: 73 with content: Hello world!
Loaded model: en-de true
elif en-de
Received message size of: 127 with content: Sticks and stones may break my bones but words WILL NEVER HURT ME!
Loaded model: en-es true
elif en-es
Received message size of: 97 with content: ¿Por qué no funciona bien?
QObject: Cannot create children for a parent that is in a different thread.
(Parent is QNetworkAccessManager(0x557d0795d7c0), parent's thread is QThread(0x557d079467e0), current thread is QThread(0x7f95604572b0)
QObject: Cannot create children for a parent that is in a different thread.
(Parent is QNetworkAccessManager(0x557d0795d7c0), parent's thread is QThread(0x557d079467e0), current thread is QThread(0x7f95604572b0)
Loaded model: es-en true
zelf en
Writing response: 90 {
"id": 0,
"success": true,
"target": {
"text": "Hallo Welt!"
}
}

Z{
"id": 0,
"success": true,
"target": {
"text": "Hallo Welt!"
}
}
Writing response: 155 {
"id": 1,
"success": true,
"target": {
"text": "Palos y piedras pueden romperme los huesos, pero las palabras NUNCA HURT ME."
}
}

{
"id": 1,
"success": true,
"target": {
"text": "Palos y piedras pueden romperme los huesos, pero las palabras NUNCA HURT ME."
}
}
Writing response: 111 {
"id": 2,
"success": true,
"target": {
"text": "Warum funktioniert es nicht gut?"
}
}

o{
"id": 2,
"success": true,
"target": {
"text": "Warum funktioniert es nicht gut?"
}
}
```
Caveats:
1) This is a horrible mess that should be cleaned up. 
2) The model book keeping code should be propagated back to the `ModelManager`. 
3) The `qEventLoop` I am using at the moment is broken because it's on the wrong thread. It needs to be fixed, but since downloading new models is not at all implemented, it doesn't matter.

Questions: How do we handle downloading a model that is not installed locally? If we just do it automatically, there would be a considerable wait between the model downloading and the translation being executed. If we block stdin reading until this is done, it will take quite a while to get translations. Not ideal.
I think that we should explicitly ask the user for a new model to be downloaded and then report when this is done. Automatically refuse translation requests for languages that we don't have? How about this? In this case, once the new model is downloaded, the client would have to re-queue everything to the server.

jerinphilip:
> I think that we should explicitly ask the user for a new model to be downloaded and then report when this is done. 

As a user, I like to be informed we're downloading a model, since downloading it will take time, it will be fast for pages after. Once done, the page translates. I will like all of these to happen automatically (only one click, my intentions are to translate, don't prompt me with n clicks) - updates should tell me progress. 

w.r.t JS Native interactions, would it a bad design to rely on translate failing with a special error, which says model not present and triggers the model download then translate workflow at extension. The catch block will do download, then when download completes automatically run the translation on the page.

XapaJIaMnu:
In the GUI, whenever there's a model downloading, the user has a visual notification of what is happening and can continue  submitting translations, because we have multiple channels of communication, so to say.

With native messaging, we avoided having a processing queue by just queuing translations to the service. Now, a request comes in with unavailable language pair. What do we do? We stall and download? As you said, bad, the user should have control over what is translated. I think the best course of action is to return error on an unavailable language pair. Now do we connect to the internet to check for available models, or do we only do that if the user prompts us? Do we send progress messages to the client when downloading or do we only do that on client request?

I think I will go and implement some combination of that and we'd see what @jelmervdl finds most comfortable to do on the front end.

jelmervdl:
How about these commands: https://gist.github.com/jelmervdl/4883a14d69d9da2f65fd60e16d94d36f

I know I said:

> In my extension I skipped the "load model" and "download & load model" messages because that's state that would need to be synced between both sides that are talking.

But I think that might have been a mistake. It's nice in theory but TranslateLocally will very likely at some point have multiple models for the same language pairs (no Jelmer, it already has! _base_ & _tiny11_!) and I'd expect some end user will want to be precise about what model to use. (I know at least one: me.)

So I added commands for listing models, and using specific models in translation requests. This way you just push the responsibility of selecting the right models to the client. The client has more opportunity to communicate with the end user, so I'd assume it would be in a better position to pick the right model. And it can deal with deciding to download models etc. better.

The `Translate` message has been updated to take model ids instead of language codes to reflect that.

Ideally I'd see the DownloadModel command implemented as a `Map<ModelID,DownloadProcess>` or something. If two messages request the same model to be downloaded, the second can just add a couple of signal handlers onto the download process the first one started already.

XapaJIaMnu:
I was also thinking of something along those lines, although I think i'd like to roll out the extra functionality in stages, not all at once.. 

I am thinking of having the main thread be responsible for consuming input and translations, but in case it encounters a message with that is concerned with model downloading, it passes it on a separate thread that does all that work (eg network access) and whenever it needs to communicate, it can try to acquire locks whenever it can and write to stdin.

In terms of the proposed message formats, several  things:

1) During a translation request the model preference should be optional. There should be a setting where the user trusts us to select the best model that we think they would like to use and that should be the default, looking at what other plugins do.
2) Instead of final, we can have 
```json
{
    ...
    part-num: 2
    part-total: 8
    ...
}
```
Then we know that this message has 8 parts and this is part 2. The vast majority of messages will be part-num 1 of 1.

3) Every message should contain status field. Either "OK" or error, at which point Error will contain text and all other fields would be undefined.

How does this sound?

XapaJIaMnu:
Current state of things.
1) Implemented the API proposed by @jelmervdl  with small modifications. I will communicate what the modifications are later today and maybe we can come to an agreement.
2) QJson doesn't offer schemas. Which sucks for validating JSONs and making sure the API is well defined.
3) Wrong Json commands sent are not at all tested.
4) Moved everything to the main thread. This allows for fetching remote models list and downloading models to work. Downloading models works but it is not fully implemented (Doesn't return proper json, doesn't handle errors, doesn't update the available model list).
5) Doesn't compile on a bunch of systems, due to not properly detecting C++17. Need to investigate (`std::variant`/`std::visit` not defined).
6) Code is messy and I would like to add extra reworks.
7) Use ../scirpts/native_client.py to quickly test the capabilities.
8) Probably something i've missed

src/cli/NativeMsgIface.cpp
```diff
@@ -295,23 +295,25 @@ request_variant NativeMsgIface::parseJsonInput(char * bytes, size_t length) {
...
 
 inline QByteArray NativeMsgIface::toJsonBytes(marian::bergamot::Response&& response, int myID) {
-    QJsonObject jsonObj;
-    jsonObj.insert(QString("id"), myID);
-    QJsonObject text;
-    text.insert(QString("text"), QString::fromStdString(response.target.text));
-    jsonObj.insert(QString("target"), text);
-    jsonObj.insert(QString("success"), true);
+    QJsonObject jsonObj{
+        {"id", myID},
```
XapaJIaMnu, line 309: Oh god, of course, initialiser lists...

XapaJIaMnu:
@jelmervdl I attempted to do a proper asynchronous  client app and so far I am failing
1) We didn't have an event loop before so we couldn't execute network operations, as they were waiting for our event loop to do signal handling. It only worked because we manually started and stopped one (`eventLoop_.exec/exit`)
2) I created a proper event loop and launched `QCoreApplication`, making the native msg iface more qtonic. https://www.qtcentre.org/threads/9678-CLI-and-QCoreApplication-where-to-start https://www.qtcentre.org/threads/41978-QCoreApplication-event-looping
3) Events are not being processed. As far as I could figure out this is because we have a while loop that is constantly doing something (or the thread is waiting for something to complete) https://www.qtcentre.org/threads/43695-Qt-Console-Application-while-loop-blocks-event-loop
4) We can force events through by issuing `QThread::currentThread()->eventDispatcher()->processEvents(QEventLoop::AllEvents)` but this is very hacky.
5) We can run the IO thread separately, but then we can't call the other QT objects'  methods as the parents are in a different thread.
6) A solution? The IO listener thread runs in its own thread and sends the input json as a signal which is later caught by the object running on the main thread and executed by its event loop. Again seems ugly, but I think this is the most qtonic. Will try it tmrw.

jelmervdl:
6 does sound like the most readable way of doing it, if you must stay inside the Qt event loop. That's necessary for the model and model list downloading mostly I think? Yeah I don't see us moving all of that to just std cpp anytime soon, and you'd probably still be stuck with an event loop for network IO, just not Qt's.

You could try to wrap all of the Qt event loop interaction into `std::async` and in that way turn the download model and download model list code into blocking threads. And then deal with them in a std::cpp fashion. But my intuition says that that only makes things more complicated, not less.

There are probably some performance implications of turning every message into a signal. But QString is a fancy shared_ptr, so sharing those across threads should be okay in terms of copying. Ideally you can go from `std::cin` directly to a `QString` without going through a `std::string` (probably via `QTextStream` or `QDataStream`…)

XapaJIaMnu:
After some reading on the internet, stdin reading should be done using `QSocketNotifier`. I am not sure if I want to get the current approach work as per 6), or try to go with `QSocketNotifier` directly.

I am somewhat concerned with the prospect of sending all of our jsons as signals. Over the weekend, I read that the doing a signal/slot call is approximately 10X more expensive than doing a simple function call, but if the bulk of the work is done translation wise, it probably wouldn't matter.

The `std::async` approach seems even more hacky to me compared to having my conditional variables execute `processEvents`...

jelmervdl:
I also found QSocketNotifier, but that would imply async IO I think? And that again would imply reading what's available and maintaining/filling a byte buffer yourself until you have a complete message. Having a thread doing blocking IO where you can just say "only return once you have N bytes for me" sounds much easier.

I was searching whether there was something like `connect(stdin.read(n_bytes), finished(QByteArray), [](QByteArray data) { ... })` in Qt but I haven't found it. Closest I found is using `QtConcurrent::run` and then having blocking IO inside whatever you're calling. But then you have to hide the while-loop inside the callback function recursively triggering the next call, which isn't any cleaner than having a separate thread with a simple blocking while loop.

Also QSocketNotifier is only useful for scheduling a "go read some bytes" call in the event loop. It doesn't do the reading, just tells the event loop that there is something readable.

XapaJIaMnu:
My understanding was that `QSocketNotifier` would get rid of the necessity to do blocking while loop waiting for io, but as we could get multiple messages, we would still have to have a while loop to process everything available on stdin..... so no.

XapaJIaMnu:
@jelmervdl  this should be ready for more extensive testing. I've changed a bit your API (model's ID is now `modelID` as opposed to `id`), but I think i have kept everything else. To test an example (from the build directory) `../scripts/native_client.py`
This results in the following stdin input:
```bash
% cat /tmp/inputtesta                                                                                                                                                                  :(
l{"command": "Translate", "id": 0, "data": {"text": "Hello world!", "src": "en", "trg": "de", "html": false}}�{"command": "Translate", "id": 1, "data": {"text": "Sticks and stones may break my bones but words WILL NEVER HURT ME!", "src": "en", "trg": "es", "html": false}}�{"command": "Translate", "id": 2, "data": {"text": "\u00bfPor qu\u00e9 no funciona bien?", "src": "es", "trg": "de", "html": false}}C{"command": "ListModels", "id": 3, "data": {"includeRemote": true}}H{"command": "DownloadModel", "id": 4, "data": {"modelID": "en-cs-tiny"}}
```
And the following output:
```bash
% cat /tmp/testa 
Ops remaining: 6
[{
    "error": "ID field in message cannot be null!",
    "id": -1,
    "success": false
}
z{
    "data": {
        "target": {
            "text": "Hallo Welt!"
        }
    },
    "id": 0,
    "success": true
}
�{
    "data": {
        "target": {
            "text": "Palos y piedras pueden romper mis huesos, pero las palabras NUNCA HURT ME."
        }
    },
    "id": 1,
    "success": true
}
�{
    "data": {
        "target": {
            "text": "Warum funktioniert es nicht gut?"
        }
    },
    "id": 2,
    "success": true
}
�{
    "data": [
        {
            "local": true,
            "modelName": "Bulgarian-English tiny",
            "repository": "Bergamot",
            "shortname": "bg-en-tiny",
            "src": "Bulgarian",
            "srcTags": {
                "bg": "Bulgarian"
            },
            "trg": "English",
            "trgTag": "en",
            "type": "tiny"
        },
        {
            "local": true,
            "modelName": "German-English tiny",
            "repository": "Bergamot",
            "shortname": "de-en-tiny",
            "src": "German",
            "srcTags": {
                "de": "German"
            },
            "trg": "English",
            "trgTag": "en",
            "type": "tiny"
        },
        {
            "local": true,
            "modelName": "English-Czech tiny",
            "repository": "Bergamot",
            "shortname": "en-cs-tiny",
            "src": "English",
            "srcTags": {
                "en": "English"
            },
            "trg": "Czech",
            "trgTag": "cs",
            "type": "tiny"
        },
        {
            "local": true,
            "modelName": "English-German tiny",
            "repository": "Bergamot",
            "shortname": "en-de-tiny",
            "src": "English",
            "srcTags": {
                "en": "English"
            },
            "trg": "German",
            "trgTag": "de",
            "type": "tiny"
        },
        {
            "local": true,
            "modelName": "English-Spanish tiny",
            "repository": "Bergamot",
            "shortname": "en-es-tiny",
            "src": "English",
            "srcTags": {
                "en": "English"
            },
            "trg": "Spanish",
            "trgTag": "es",
            "type": "tiny"
        },
        {
            "local": true,
            "modelName": "English-Finnish npz",
            "repository": "unknown",
            "shortname": "eng-fin-npz",
            "src": "English",
            "srcTags": {
            },
            "trg": "Finnish",
            "trgTag": "",
            "type": "transformer-tiny11-npz"
        },
        {
            "local": true,
            "modelName": "English-Finnish tiny",
            "repository": "unknown",
            "shortname": "eng-fin-tiny",
            "src": "English",
            "srcTags": {
            },
            "trg": "Finnish",
            "trgTag": "",
            "type": "transformer-tiny11"
        },
        {
            "local": true,
            "modelName": "Spanish-English tiny",
            "repository": "Bergamot",
            "shortname": "es-en-tiny",
            "src": "Spanish",
            "srcTags": {
                "es": "Spanish"
            },
            "trg": "English",
            "trgTag": "en",
            "type": "tiny"
        },
        {
            "local": true,
            "modelName": "Swedish-Finnish tiny",
            "repository": "unknown",
            "shortname": "swe-fin-tiny",
            "src": "Swedish",
            "srcTags": {
            },
            "trg": "Finnish",
            "trgTag": "",
            "type": "transformer-tiny11"
        },
        {
            "local": false,
            "modelName": "Czech-English base",
            "repository": "Bergamot",
            "shortname": "cs-en-base",
            "src": "Czech",
            "srcTags": {
                "cs": "Czech"
            },
            "trg": "English",
            "trgTag": "en",
            "type": "base"
        },
        {
            "local": false,
            "modelName": "Czech-English tiny",
            "repository": "Bergamot",
            "shortname": "cs-en-tiny",
            "src": "Czech",
            "srcTags": {
                "cs": "Czech"
            },
            "trg": "English",
            "trgTag": "en",
            "type": "tiny"
        },
        {
            "local": false,
            "modelName": "German-English base",
            "repository": "Bergamot",
            "shortname": "de-en-base",
            "src": "German",
            "srcTags": {
                "de": "German"
            },
            "trg": "English",
            "trgTag": "en",
            "type": "base"
        },
        {
            "local": false,
            "modelName": "English-Bulgarian tiny",
            "repository": "Bergamot",
            "shortname": "en-bg-tiny",
            "src": "English",
            "srcTags": {
                "en": "English"
            },
            "trg": "Bulgarian",
            "trgTag": "bg",
            "type": "tiny"
        },
        {
            "local": false,
            "modelName": "English-Czech base",
            "repository": "Bergamot",
            "shortname": "en-cs-base",
            "src": "English",
            "srcTags": {
                "en": "English"
            },
            "trg": "Czech",
            "trgTag": "cs",
            "type": "base"
        },
        {
            "local": false,
            "modelName": "English-German base",
            "repository": "Bergamot",
            "shortname": "en-de-base",
            "src": "English",
            "srcTags": {
                "en": "English"
            },
            "trg": "German",
            "trgTag": "de",
            "type": "base"
        },
        {
            "local": false,
            "modelName": "English-Estonian tiny",
            "repository": "Bergamot",
            "shortname": "en-et-tiny",
            "src": "English",
            "srcTags": {
                "en": "English"
            },
            "trg": "Estonian",
            "trgTag": "et",
            "type": "tiny"
        },
        {
            "local": false,
            "modelName": "Estonian-English tiny",
            "repository": "Bergamot",
            "shortname": "et-en-tiny",
            "src": "Estonian",
            "srcTags": {
                "et": "Estonian"
            },
            "trg": "English",
            "trgTag": "en",
            "type": "tiny"
        },
        {
            "local": false,
            "modelName": "Icelandic-English tiny",
            "repository": "Bergamot",
            "shortname": "is-en-tiny",
            "src": "Icelandic",
            "srcTags": {
                "is": "Icelandic"
            },
            "trg": "English",
            "trgTag": "en",
            "type": "tiny"
        },
        {
            "local": false,
            "modelName": "Norwegian (Bokmål)-English tiny",
            "repository": "Bergamot",
            "shortname": "nb-en-tiny",
            "src": "Norwegian (Bokmål)",
            "srcTags": {
                "nb": "Norwegian (Bokmål)"
            },
            "trg": "English",
            "trgTag": "en",
            "type": "tiny"
        },
        {
            "local": false,
            "modelName": "Norwegian (Nynorsk)-English tiny",
            "repository": "Bergamot",
            "shortname": "nn-en-tiny",
            "src": "Norwegian (Nynorsk)",
            "srcTags": {
                "nn": "Norwegian (Nynorsk)"
            },
            "trg": "English",
            "trgTag": "en",
            "type": "tiny"
        }
    ],
    "id": 3,
    "success": true
}
Ops remaining: 1
&{
    "data": [
        {
            "progress": 0.001099014174089909
        },
        {
            "url": "https://data.statmt.org/bergamot/models/csen/encs.student.tiny11.tar.gz"
        },
        {
            "modelID": "en-cs-tiny"
        }
    ],
    "id": 4,
    "success": true
}
Ops remaining: 1
Ops remaining: 1
Ops remaining: 1
%{
    "data": [
        {
            "progress": 0.21320874977344237
        },
        {
            "url": "https://data.statmt.org/bergamot/models/csen/encs.student.tiny11.tar.gz"
        },
        {
            "modelID": "en-cs-tiny"
        }
    ],
    "id": 4,
    "success": true
}
Ops remaining: 1
${
    "data": [
        {
            "progress": 0.4286155278950645
        },
        {
            "url": "https://data.statmt.org/bergamot/models/csen/encs.student.tiny11.tar.gz"
        },
        {
            "modelID": "en-cs-tiny"
        }
    ],
    "id": 4,
    "success": true
}
${
    "data": [
        {
            "progress": 0.6143489233162591
        },
        {
            "url": "https://data.statmt.org/bergamot/models/csen/encs.student.tiny11.tar.gz"
        },
        {
            "modelID": "en-cs-tiny"
        }
    ],
    "id": 4,
    "success": true
}
Ops remaining: 1
${
    "data": [
        {
            "progress": 0.8088744321301731
        },
        {
            "url": "https://data.statmt.org/bergamot/models/csen/encs.student.tiny11.tar.gz"
        },
        {
            "modelID": "en-cs-tiny"
        }
    ],
    "id": 4,
    "success": true
}
Ops remaining: 1
{
    "data": [
        {
            "progress": 1
        },
        {
            "url": "https://data.statmt.org/bergamot/models/csen/encs.student.tiny11.tar.gz"
        },
        {
            "modelID": "en-cs-tiny"
        }
    ],
    "id": 4,
    "success": true
}
r{
    "data": [
        {
            "modelID": "en-cs-tiny"
        }
    ],
    "id": 4,
    "success": true
}
```
Or alternatively you can run like: `cat /tmp/inputtesta | ./translateLocally -p`

Caveats:
1) I haven't checked if I have implemented all the API correctly (there are several schema libraries, maybe a good idea to use one.)
2) Termination condition: Either `stdin` is closed, or the server receives a message of length `-1`.
3) I have a while loop that waits for all events to terminate and some complex and error prone counter mechanism, that aims to keep track of all operations. I would really like to think of a better way around it.
4) It doesn't compile, due to me using c++17 features. Should investigate if this is a bug in CMake and we're given an older standard, or some compilers genuinely do not support c++17 from some of our runs.

jelmervdl:
I removed the `success:true` bit from the download progress updates. This so I can keep the [code that handles incoming messages](https://github.com/jelmervdl/firefox-translations/blob/ebe33ab6fb231b0cfbb1f50e15fca17641d592d2/extension/controller/translation/translationHelper.js#L71-L89) simple and generic. I also changed the contents of `data` to be just an object with keys instead of the odd array of objects with each one key/value pair. I assumed that is was an array by 

I'll look into 2 and 3 as well. I can't think of an easy solution for those. I don't know how Firefox signals the native app to stop if the extension disconnects or is reloaded (but that [should not be too difficult to figure out](https://searchfox.org/mozilla-central/source/toolkit/components/extensions/NativeMessaging.jsm#54)). Ideally it would be something like you're already implementing, stdin stops new requests from coming in but old events are still handled. A counter would indeed be a simple solution for that, but also error prone? Client app can also opt to just sigkill translateLocally if they want to quit quickly (no need to wait around for translations to finish if there isn't anyone listening on the other side. Model downloads in the progress of extraction though might be troubling?)

XapaJIaMnu:
I think I wasn't clear in my post. Regarding termination, 2 is already implemented but possibly can be improved.

I also think that the behaviour I have implemented for shutdown (wait until all operations are finished before deleting) is useful for offline testing without an extension more so than for the actual extension

jelmervdl:
The translation requests don't work anymore for me. I always get _Failed to load the necessary translation models._

Sending:
```json
{
  "id": 53,
  "command": "Translate",
  "data": {
    "src": "de",
    "trg": "en",
    "text": "\n\t\t<span class=\"vector-menu-heading-label\" data-x-bergamot-id=\"0\">Namensräume</span>\n\t",
    "html": true
  }
}
```

I do have a model for it:
```
> ll
Permissions Size User   Group Date Modified Name
drwxr-xr-x@    - jelmer staff  3 Mar 18:29  bgen.student.tiny11-1646332153
drwxr-xr-x@    - jelmer staff  3 Mar 18:28  csen.student.base-1646332084
drwxr-xr-x@    - jelmer staff  3 Mar 18:30  deen.student.base-1646332215
drwxr-xr-x     - jelmer staff  4 Mar 10:58  encs.student.tiny11-1646391519
drwxr-xr-x@    - jelmer staff  3 Mar 18:35  isen.student.tiny11-1646332507
drwxr-xr-x@    - jelmer staff  3 Mar 18:33  nnen.student.tiny11-1646332436
```

And when running translateLocally in GUI mode, de->en translation works as expected.

There is also a bit of a design problem: The `Translate` request interface, with only `src` and `trg`, has the issue that you cannot specify which model(s) to use. We already have multiple models for language pairs (base, tiny11) and I assume that will only increase when more repositories start showing up. The `ListModels` request returns all models, but there is no way to distinguish between many of them. The option to pivot makes the number of models that could match a `src` and `trg` selection even larger.

XapaJIaMnu:
I think you might have the old version of the models that do not have a tag in the . Delete them and  try to redownload?

As for model preference, I'm not entirely sure how to encode that, but it shouldn't be too much of an issue. I can code something to that end next week.

jelmervdl:
I updated the python code to be a bit more like a tiny library that we can do little tests with.

I started doing some timing measurements of the interface. If you give it enough to do, responses will take a while. No surprises there. But all in all it is already way way faster than WASM at the moment. (I don't have any actual measurements of wasm but it all feels pretty fast.) [measurements.csv](https://github.com/XapaJIaMnu/translateLocally/files/8236325/measurements.csv), [Notebook with scatterplot](https://colab.research.google.com/drive/1rqbh5PMGyDK-VZWG2WhJOzrFeiHBabuN?usp=sharing)

It also seems to be reasonably stable. Thinking about what is left:
- [ ] Still need to fix old macOS understanding this is C++17. 
- [ ] I removed some of your optimisations, such as the map for source/target models (its just a linear scan) and the throttled download progress message. Qt already throttles this to a degree, and I don't think it in any way taxes the native messaging channel. So I'm not sure whether we need to bring it back. This also gives you a much smoother progress bar, for those who use the cursor to see whether it moved a pixel ;)
- [ ] Other possible optimisation: keeping the loaded models around for a bit longer (e.g. when you're switching between two translating browser tabs in different source languages, the model now reloads the model each time when switching)
- [ ] Oh and I want to add selecting translation models by their id instead of just src and trg language tags. That way you can be very specific which model (and optionally which pivot) you're using. For those who want to compare multiple models with the same src and target language. I saw you already had `chosemodel` for it. I'm tempted to just go with `model` and optionally `pivot` (as alternatives to `src` and `trg`) (or, to stay in line with the `DownloadModel` terminology, `modelID` and `pivotModelID`? But that's so verbose.)
- [ ] Alignments aren't supported yet. We should export those as well (although we have no consumer for it.). I think we can ignore quality output for the time being.
- [ ] Other todos seem mostly related to graceful error handling and cleanup I think.

src/cli/NativeMsgIface.cpp
```diff
@@ -0,0 +1,454 @@
...
+                {"success", true},
+                {"id", myJsonInput.id},
+                {"data", QJsonObject{{"modelID", model->id()}}}
+            };
+            lockAndWriteJsonHelper(QJsonDocument(jsonObj).toJson());
+            return;
+        }
+
+        // Download new model
+        QNetworkReply *reply = network_.downloadFile(model->url, QCryptographicHash::Sha256, model->checksum, QVariant::fromValue(QMap<QString, QVariant>({
```
XapaJIaMnu, line 245: Shouldn't we check if the reply is valid here and emit an error if it isn't?

src/cli/NativeMsgIface.cpp
```diff
@@ -0,0 +1,454 @@
...
+            {"id", myJsonInput.id},
+            {"modelID", model->id()}
+        })));
+
+        // Pass any download progress updates along to the client. Previous implementation
+        // has some logic to limit the amount of progress updates to 6 but I don't
+        // think throttling the number of messages is necessary.
+        connect(reply, &QNetworkReply::downloadProgress, this, [=](qint64 ist, qint64 max) {
```
XapaJIaMnu, line 253: Is this connection bound to the `reply` object, and therefore no need to uniquify-it?

src/cli/NativeMsgIface.cpp
```diff
@@ -0,0 +1,454 @@
...
+    // Connection to store our callback in. Qt6 has single shot connections, but 5 doesn't.
+    QMetaObject::Connection * const connection = new QMetaObject::Connection;
+
+    auto downloadModelLambda = std::function([=](){
+        // Remove the connection right after we fetch model, if the connection is valid
+        if (*connection)
+            QObject::disconnect(*connection);
+
+        // Clean up our callback connection
+        // TODO: abstract this all away in a connectSingleShot() helper?
```
XapaJIaMnu, line 222: Maybe? If we do that though, we should use QT6 singleShot connections so that we can readily remove the qt5 codepath in... 5 years when the oldest ubuntu LTS ships with QT6.....

src/cli/NativeMsgIface.cpp
```diff
@@ -0,0 +1,454 @@
...
+
+        // Download new model
+        QNetworkReply *reply = network_.downloadFile(model->url, QCryptographicHash::Sha256, model->checksum, QVariant::fromValue(QMap<QString, QVariant>({
+            {"id", myJsonInput.id},
+            {"modelID", model->id()}
+        })));
+
+        // Pass any download progress updates along to the client. Previous implementation
+        // has some logic to limit the amount of progress updates to 6 but I don't
+        // think throttling the number of messages is necessary.
```
XapaJIaMnu, line 252: My reasoning with regards to throttling is that given a very slow network, the client would be spammed forever with incremental progress messages, though i am not sure if in practise this would mean anything much.

XapaJIaMnu (commented):
>> Old compilers and C++17
This is also an issue on old ubuntu and Windows.
>> Better model search and precise model selection
We should refactor `ModelManager` to contain the models in a convenient format where we can query them by tag, as well as by ID. This will also allow the main client to pivot.

To verify: Can we confirm that tiny, basic and 3rd party models load fine with the dynamic model swapping? If not we should conditionally reload the service whenever the architecture has changed.

src/cli/NativeMsgIface.cpp
jelmervdl: Exactly!

src/cli/NativeMsgIface.cpp
jelmervdl: You're right, reply could be nullptr if it fails to open a tempfile. I'll fix it.

jelmervdl:
I think I have messed up the `operations_--` counter somewhere as well. I'm thinking about maybe replacing it with something based on RAII, maybe I can make something that uses move semantics on the `request_variant` which can then be passed around, like a shared or unique ptr. Then, as long as there are still request_variants alive, don't stop the event loop.

jelmervdl:
> To verify: Can we confirm that tiny, basic and 3rd party models load fine with the dynamic model swapping? If not we should conditionally reload the service whenever the architecture has changed.

I've tested this by switching rapidly between the en-de-tiny, en-de-base, eng-fin-tiny and eng-ukr-tiny models and they all seem to produce translations that seem plausible.

jelmervdl:
I was afraid that you could have in theory a scenario where things would go out of sync:

```cpp
// worker thread:
operations_--     // 1
cv.notify_one();  // 2

// iothread
cv.wait(lock, [this]() { return operations_ == 0; });  // 3
```

Basically, my understanding of atomic was that, _3_ could still be reading the old value from before _1_ because, while incrementing and decrementing are guaranteed to be atomic, reading did not have to be. `shared_ptr`'s `use_count()` seems to have this behaviour. But I was wrong.

If I understand [`memory_order_seq_cst`](https://en.cppreference.com/w/cpp/atomic/memory_order) correctly, reading from one thread after updating it from another (and the *after* is guaranteed by the condition_variable notification) should be fine.

So I reverted it back to using atomic.

src/cli/NativeMsgIface.cpp
```diff
@@ -0,0 +1,430 @@
...
+    std::ios_base::sync_with_stdio(false);
+
+    // Init the marian translation service:
+    marian::bergamot::AsyncService::Config serviceConfig;
+    serviceConfig.numWorkers = settings_.marianSettings().cpu_threads;
+    serviceConfig.cacheSize = settings_.marianSettings().translation_cache ? kTranslationCacheSize : 0;
+    service_ = std::make_shared<marian::bergamot::AsyncService>(serviceConfig);
+
+    // Pick up on network errors
+    // TODO: not sure if every request always returns on 1 error. fetchRemoteModels for example
```
XapaJIaMnu, line 88: Should we have some protection so that no parallel downloads are executed? Since we increment `operations_` before the download request is sent, we would end up with mismatched increment as no error is sent back and no way to decrement the counter?

src/cli/NativeMsgIface.cpp
```diff
@@ -0,0 +1,430 @@
...
+    // will cut short if a request is already in progress. If that one fails, the second request
+    // that got cut short will have no response, and stay pending.
+    connect(&network_, &Network::error, this, [&](QString err, QVariant data) {
+        if (data.canConvert<Request>())
```
XapaJIaMnu, line 92: I think we should have an else branch that writes to stderr. The else branch is a programmer error in this case.

src/cli/NativeMsgIface.cpp
```diff
@@ -0,0 +1,430 @@
...
+            writeError(data.value<Request>(), std::move(err));
+    });
+
+    connect(&network_, &Network::downloadComplete, this, [this](QFile *file, QString filename, QVariant data) {
+        models_.writeModel(file, filename);
+
+        if (data.canConvert<DownloadRequest>()) {
```
XapaJIaMnu, line 99: Else branch, as per above comment.

src/cli/NativeMsgIface.cpp
```diff
@@ -0,0 +1,430 @@
...
+void NativeMsgIface::run() {
+    iothread_ = std::thread([this](){
+        for (;;) {
+            // First part of the message: Find how long the input is. If that
+            // read fails, we're probably at EOF.
+            char len[4];
+            if (!std::cin.read(len, 4))
+                break;
+
+            int ilen = *reinterpret_cast<unsigned int *>(len);
```
XapaJIaMnu, line 117: currently `_die` is never read, so we should probably remove it. I think That we can use `ilen == -1` as a signal that we want the native message client to die cleanly, as opposed to closing the `cin`.

src/cli/NativeMsgIface.cpp
```diff
@@ -0,0 +1,430 @@
...
+            if (ilen >= kMaxInputLength || ilen < 2) { // >= 2 because JSON is at least "{}"
+                std::cerr << "Invalid message size. Shutting down." << std::endl;
+                break;
+            }
+
+            //  Read in the message into Json
+            std::shared_ptr<std::vector<char>> input(std::make_shared<std::vector<char>>(ilen));
```
XapaJIaMnu, line 124: What's the benefit of using `shared_ptr` here, if we are already using the heavyweight vector? I think `std::array` would be a better choice, since the input is constant size?

XapaJIaMnu (commented):
I did a first round review. `NativeMsgIface.cpp` could benefit from another round. Mostly small comments. The bigger  change that could be done is to make `NativeMsgIface` extend `QThread`. That might  make the code a bit more readable.

src/cli/CLIParsing.h
```diff
@@ -22,25 +28,38 @@ static void CLIArgumentInit(QAppType& translateLocallyApp, QCommandLineParser& p
...
         if (parser.isSet(flag)) {
-            cliONLY = true;
-            break;
+            return CLI;
         }
     }
-    return cliONLY;
+
+    for (auto&& path : parser.positionalArguments()) {
+        if (path == "{c9cdf885-0431-4eed-8e18-967b1758c951}") {
```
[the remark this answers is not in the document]
jelmervdl: It searches for the browser extension ID in the arguments. That seems to be the common thing in how Chrome and Firefox launch applications through their native messaging API:

> Two arguments are passed to the native app when it starts:
> - The complete path to the app manifest.
> - (new in Firefox 55) the ID (as given in the browser_specific_settings manifest.json key) of the add-on that started it.
>
> Note: Chrome handles the passed arguments differently:
> - On Linux and Mac, Chrome passes one argument: the origin of the extension that started it (in the form chrome-extension://[extensionID]). This enables the app to identify the extension.
> - On Windows, Chrome passes two arguments: the first is the origin of the extension, and the second is a handle to the Chrome native window that started the app.

Source: https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/Native_messaging#extension_side

src/cli/NativeMsgIface.cpp
jelmervdl: I checked parallel downloads of different models. That is a safe operation and works. All code with `fetchRemoteModels` should also be okay, I checked that as well. I haven't checked multiple downloads of the same model. I think in that case it would successfully download the model multiple times (so you end up with multiple copies on disk, of which only one is used and listed) but the number of operations would still match up.

src/mainwindow.cpp
```diff
@@ -66,14 +67,21 @@ MainWindow::MainWindow(QWidget *parent)
...
+    }
 
     // If no model is preferred, load the first available one.
     if (settings_.translationModel().isEmpty() && !models_.getInstalledModels().empty())
         settings_.translationModel.setValue(models_.getInstalledModels().at(0).path);
 
     // Attach slots
+    
+    // As soon as we've started up, try to register this application as a native messaging host
+    connect(this, &MainWindow::launched, this, &MainWindow::registerNativeMessagingAppManifest);
```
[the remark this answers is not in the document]
jelmervdl: Bit more context: browser plugins (which I assume is the mayor use case for native messaging clients) are not privileged to register native messaging apps. Apps like 1Password ship as an installer for this reason I expect.

src/cli/NativeMsgIface.cpp
jelmervdl: The shared_ptr is to avoid copies as the input goes through Qt's signal/slot interface. The vector because macOS 10.15 ships with a compiler that doesn't like smart pointers and C-style arrays for some reason. `std::array` won't work here because the size of that thing is a template variable, not a runtime variable.

But I just replaced all of it with a QByteArray, which was the type it will be converted to anyway! And QByteArray also does all the copy-on-write Qt smart copy stuff in the background, so no need for a shared_ptr :)

src/mainwindow.cpp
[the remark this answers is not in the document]
kpu: So longer term we need an installer (especially on Windows) but for now we are abusing the GUI application as an installer.  Makes some sense.

src/cli/NativeMsgIface.cpp
jelmervdl: Closing cin is the common way to signal a downstream process to finish up? It is how pipes work. The native message server will stop only after it finished handling all pending requests after its stdin was closed. I have a test in `native_client.py` to test for this.

src/cli/NativeMsgIface.h
```diff
@@ -0,0 +1,269 @@
...
+    class AsyncService;
+    class TranslationModel;
+    class Response;
+    }
+}
+
+
+const int constexpr kMaxInputLength = 10*1024*1024; // 10 MB limit on the input length via native messaging
+
+struct Request {
```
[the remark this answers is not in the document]
jelmervdl: I didn't read the "*I* want to add" so I went through and added the expected JSON to each of the request types 😅

jelmervdl:
This code should now work with [the latest version](https://github.com/jelmervdl/firefox-translations/releases/tag/v0.4.1) of my Firefox plugin. Be sure to switch to TranslateLocally in the extension settings.

(I have only tested it on macOS so far, but will test it with Windows and Ubuntu when I have the time for those VMs to boot up, install Firefox in them, etc.)

jelmervdl:
I fixed the binary/text mode conversion on Windows. The python script now runs through its tests without issue 🥳 

The debug build on windows runs into another issue though, that I have no clue about… I'll wait for Github CI to build a release build to see whether that does work. But if anyone has seen anything like this before, let me know:
image

jelmervdl:
Okay, found that as well. Waiting for https://github.com/browsermt/bergamot-translator/pull/396 to be merged, and then I'll pull that into one final commit to get browser translations on Windows to work out of the box.

jelmervdl:
I've confirmed that native client now works with Firefox and the [browser extension](https://github.com/jelmervdl/firefox-translations) out of the box (well, you have to start TranslateLocally once) in Windows 10, macOS 12.3 and Ubuntu 20.10. 🎊 

image

XapaJIaMnu:
We should add the native message configuration to the read me.

XapaJIaMnu:
@jelmervdl , could you please add to the README.md instructions how to use the native message interface, and then we can merge this.

jelmervdl:
I added [a brief description about the native messaging functionality](https://github.com/XapaJIaMnu/translateLocally/tree/nativemsg_cli#nativemessaging-interface). It is mostly just pointers to where to find the information you'd need if you want to develop for it. Let me know whether you think anything is missing.

XapaJIaMnu:
I think this is fine for now. We can do further prs like the one that would fix #99 later on. I'm in favour of merging it.

PR #95, opened 2022-02-16
Add support for langid tags and some code cleanup

XapaJIaMnu:
Wrapped one string in tr, removed unneeded headers and forward defines and used constexpr if template instead of overloaded functions


XapaJIaMnu:
Superseded by #93

PR #101, opened 2022-04-27
Register native messaging capabilities with Chrome and Chromium

jelmervdl:


src/mainwindow.cpp
```diff
@@ -524,38 +524,60 @@ bool MainWindow::registerNativeMessagingAppManifest() {
...
-    manifestFile.open(QFile::WriteOnly);
-    manifestFile.write(manifest.toJson());
+#if defined (Q_OS_WIN)
+    QStringList registryKeys;
+    registryKeys << QString("HKEY_CURRENT_USER\\Software\\Mozilla\\NativeMessagingHosts\\%1").arg(name);
+    registryKeys << QString("HKEY_CURRENT_USER\\Software\\Google\\Chrome\\NativeMessagingHosts\\%1").arg(name);
+    registryKeys << QString("HKEY_CURRENT_USER\\Software\\Chromium\\NativeMessagingHosts\\%1").arg(name);
+
+    for (QString const &key : registryKeys)
+        QSettings(key, QSettings::NativeFormat).setValue("Default", manifestPath);
```
XapaJIaMnu, line 579: You meant for this to be inside the for loop above?

jelmervdl: For Windows I only write the manifest file itself only once and then register all the different registry keys to point to the same file.

But you're right that it is weird to see `manifestPath` outside of the for-loop. Its probably safer to just move all of this inside the for-loop since `manifestPaths` is `size() == 1` anyway in Win.

XapaJIaMnu: Well, this is why it doesn't compile on windows, but yeah I'd say it should be in the for loop for clarity's sake.

jelmervdl: Oh. Ooops 😬

XapaJIaMnu:
Only comment is that I think all the nativeMsgIface registration stuff should ideally go into a separate source file and not be sprinkled in mainwindow. Just call one function that does all from mainwindow? What do you think?

jelmervdl:
I agree, it should be a separate bit of code. I'd also like to be able to call it from the CLI, to have an option to add extension ids (so you can register your extension while developing it) and to register without having to completely start translateLocally (say from an installation script)

src/cli/CLIParsing.h
```diff
@@ -57,12 +62,31 @@ static AppType runType(QCommandLineParser& parser) {
...
+        }
+
+        // Matching anything Chromium based
+        auto match = re.match(path);
+        if (match.hasMatch() && settings.nativeMessagingClients().contains(match.captured(1))) {
             return NativeMsg;
         }
     }
 
+    // TODO: if this program is started as a browser's native messaging client,
```
XapaJIaMnu, line 83: Have you tested this, or are we leaving it as a TODO

src/cli/CommandLineIface.cpp
```diff
@@ -245,3 +254,48 @@ void CommandLineIface::outputTranslation(Translation output) {
...
 }
+
+int CommandLineIface::allowNativeMessagingClient(QStringList ids) {
+    if (ids.isEmpty()) {
+        qCritical().noquote() << "No client ids specified";
+        return 1;
+    }
+
+    auto clients = settings_.nativeMessagingClients();
+#if (QT_VERSION < QT_VERSION_CHECK(5, 14, 0))
```
XapaJIaMnu, line 265: Can we unify the codepath here? Or is the OLD qt version way of doing things deprecated?

src/cli/CommandLineIface.cpp
```diff
@@ -245,3 +254,48 @@ void CommandLineIface::outputTranslation(Translation output) {
...
+
+int CommandLineIface::removeNativeMessagingClient(QStringList ids) {
+    if (ids.isEmpty()) {
+        qCritical().noquote() << "No client ids specified";
+        return 1;
+    }
+
+    auto clients = settings_.nativeMessagingClients();
+#if (QT_VERSION < QT_VERSION_CHECK(5, 14, 0))
+    clients += ids.toSet();
```
XapaJIaMnu, line 282: This seems to be wrong. You probably want `-=` here. I think you may have accidentally copy pasted it?

src/cli/CommandLineIface.cpp
```diff
@@ -245,3 +254,48 @@ void CommandLineIface::outputTranslation(Translation output) {
...
+    auto clients = settings_.nativeMessagingClients();
+    for (auto &&client : clients)
+        out << client << "\n";
+    return 0;
+}
+
+int CommandLineIface::updateNativeMessagingManifests() {
+    NativeMsgManager manager;
+    return manager.writeNativeMessagingAppManifests(settings_.nativeMessagingClients()) ? 0 : 1;
+}
```
XapaJIaMnu, line 301: Newline

src/cli/NativeMsgManager.cpp
```diff
@@ -0,0 +1,115 @@
+#include "NativeMsgManager.h"
+#include <QCoreApplication>
+#include <QDebug>
```
XapaJIaMnu, line 3: Xtra include?

src/cli/NativeMsgManager.cpp
```diff
@@ -0,0 +1,115 @@
...
+#include <QList>
+#include <QPair>
+#include <QRegularExpression>
+#include <QSet>
+#include <QSettings>
+#include <QStandardPaths>
+#include <QStringList>
+
+
+static QRegularExpression firefoxExtensionPattern("^\\{[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}\\}$|^[a-z0-9-._]*@[a-z0-9-._]+$", QRegularExpression::CaseInsensitiveOption);
```
XapaJIaMnu, line 20: WHy not one for Chromium? Could you add a comment explaining this regex.

src/cli/NativeMsgManager.cpp
```diff
@@ -0,0 +1,115 @@
...
+#elif defined (Q_OS_LINUX)
+    manifestPaths << qMakePair(Firefox,  QString("%1/.mozilla/native-messaging-hosts/%2.json").arg(QStandardPaths::writableLocation(QStandardPaths::HomeLocation)).arg(name));
+    manifestPaths << qMakePair(Chromium, QString("%1/.config/google-chrome/NativeMessagingHosts/%2.json").arg(QStandardPaths::writableLocation(QStandardPaths::HomeLocation)).arg(name));
+    manifestPaths << qMakePair(Chromium, QString("%1/.config/chromium/NativeMessagingHosts/%2.json").arg(QStandardPaths::writableLocation(QStandardPaths::HomeLocation)).arg(name));
+#elif defined (Q_OS_WIN)
+    // On Windows, we write the manifest to some safe directory, and then point to it from the Registry.
+    manifestPaths << qMakePair(Firefox,  QString("%1/%2-firefox.json").arg(QStandardPaths::writableLocation(QStandardPaths::AppConfigLocation)).arg(name));
+    manifestPaths << qMakePair(Chromium, QString("%1/%2-chromium.json").arg(QStandardPaths::writableLocation(QStandardPaths::AppConfigLocation)).arg(name));
```
XapaJIaMnu, line 74: No chrome manifest for windows?

src/settings/Settings.h
```diff
@@ -47,60 +49,49 @@ class Setting : public QObject {
...
+        emitValueChanged(name_, backing_.value(name_)); // not using value() because we *want* the QVariant
     }
 
-    void appendToValue(QStringList entry, Behavior behavior = EmitWhenChanged) {
-        if  constexpr (std::is_same_v<QList<QStringList>, T>) {
-            value_.append(entry);
-            emitValueChanged(name_, QVariant::fromValue<QList<QStringList>>(value_));
-        } else {
-            Q_ASSERT_X(true, "appendToValue", (std::string("Developer error: Attempted to append a setting entry to a non-appendable type: ") + std::string(typeid(T).name())).c_str());
+    // Here for compatibility. Will be removed once #102 is merged.
```
XapaJIaMnu, line 76: Should we merge #102 beforehand then?

XapaJIaMnu (commented):
A few questions and suggestions. Main question: Should we merge #102 first?

src/cli/CommandLineIface.cpp
jelmervdl: `toSet()` has disappeared completely from the documentation so I don't expect it to work anymore. The canonical way of doing it is through the iterator initialisers but those are only available since 5.14 which is somehow not what Ubuntu 18.04 ships with.

src/cli/CLIParsing.h
jelmervdl: Good catch. I haven't tried it, will do so before I merge.

src/cli/CommandLineIface.cpp
jelmervdl: Yes! Nice catch! Thanks!

src/cli/NativeMsgManager.cpp
jelmervdl: I needed QDebug for `qWarning()` on Windows

src/cli/NativeMsgManager.cpp
jelmervdl: I'll add a comment. It's described a little later on as well that Firefox only allows "valid" extension ids in the manifest file otherwise it will fail to be parsed.

src/cli/NativeMsgManager.cpp
jelmervdl: Chromium and Chrome share the same file (but different registry keys, mentioned later on in the file.)

src/settings/Settings.h
jelmervdl: That would be the proper way of doing it I suppose :)

jelmervdl:
I just started checking this, but doesn't look great:

```
> ./translateLocally --list-clients
fish: Job 1, './translateLocally --list-clien…' terminated by signal SIGSEGV (Address boundary error)
...
* thread #1, queue = 'com.apple.main-thread', stop reason = EXC_BAD_ACCESS (code=1, address=0x8)
  * frame #0: 0x0000000104e237ec QtCore`QSettings::value(QString const&, QVariant const&) const + 44
    frame #1: 0x000000010016b3b9 translateLocally`SettingImpl<QList<QList<QString> > >::value() const + 73
    frame #2: 0x00000001001601d3 translateLocally`SettingImpl<QList<QList<QString> > >::operator()() const + 35
    frame #3: 0x000000010015fe39 translateLocally`Settings::repos() const + 217
    frame #4: 0x000000010013f62d translateLocally`ModelManager::ModelManager(QObject*, Settings*) + 269
    frame #5: 0x000000010013fee5 translateLocally`ModelManager::ModelManager(QObject*, Settings*) + 37
    frame #6: 0x00000001000eac39 translateLocally`CommandLineIface::CommandLineIface(QObject*) + 89
    frame #7: 0x00000001000eb66d translateLocally`CommandLineIface::CommandLineIface(QObject*) + 29
    frame #8: 0x000000010001412c translateLocally`main + 860
    frame #9: 0x000000010278152e dyld`start + 462
```

No idea why yet though. The merge with master doesn't change anything around that bit as far as I can tell.

jelmervdl:
Another thing I haven't tested, but I fear that all of this work won't work in the sandboxed version of the app, because it won't be able to see nor write to the native messaging host directories 😭 .

I think there is a way to work around it, you can trigger something to let the app ask for full disk access, but I haven't figured out yet how. Need to do a bit more testing on that.

jelmervdl:
Nvm found the issue in the difference in order of how the member variables in CommandLineIface.{h,cpp} are initialised.

src/cli/CommandLineIface.h
```diff
@@ -15,16 +15,15 @@
...
-    ModelManager models_;
-    
+    // Event loop that would wait until translation completes
+    QEventLoop eventLoop_;
+
     // Settings, network and translator:
-    Settings settings_;
     Network network_;
+    Settings settings_;
+    ModelManager models_;
```
XapaJIaMnu, line 24: I'm so confused here. In the `CommandLineIface.cpp` we have:
```cpp
, models_(this, &settings_)
```
Models requires settings, but now we have some (inversed?!?!) order dependency? That sounds very very wrong.

jelmervdl: Looks pretty normal to me. ModelManager needs settings, but isn't the only one that needs settings. You don't want modelmanager to manage its own separate copy, nor do you want to instantiate modelmanager to get access to settings. The only thing that makes it look weirder is that both modelmanager and settings are member variables.

jelmervdl: But to be fair it does look out of place in Qt context. I think Qt solves this by having all QSettings instances share the same backend and sync them under the hood. We could piggyback on that synchronisation.

XapaJIaMnu: From here it makes it seem that `settings_` has a hidden dependencies on model manager and can't be used separately.

jelmervdl: I don't think I follow… Settings.h has no reference to `ModelManager`, and can function without it.

Is it because we pass in a pointer to `settings_` instead of a reference when constructing `models_`? The only argument for that is that it allows you to do `new ModelManager(...)` and have it take ownership of the Settings instance it uses by calling `setParent()`. Like it has ownership of its own Network object. It seems to be the Qt way. But if it was just std C++, I would have passed it by reference. (If only C++ had explicit lifetimes in its typesystem…)

> We could piggyback on that synchronisation.
I was wrong. QSettings does not emit anything when a value is changed, so there is a reason we need to share one Settings instance throughout the application. `Settings` could be singleton, but I generally don't like those.

XapaJIaMnu: Sorry, my bad. I misunderstood that models need to get initialised before settings, but yeah it's the other way around of course. Really my bad, sorry.

PR #102, opened 2022-04-28
Model management improvements

jelmervdl:
I'm adding model metadata to a separate file when you download a model. This is stuff the model itself might not know about it (or might lie about) such as when and where it was downloaded from.

I'm storing it as a separate file, and not inside model_info.json, so we don't have to edit that file from inside translateLocally (i.e. not running the risk losing json data that that particular version of translateLocally doesn't recognise) and we can just change the file format whenever we like it.

Todo:
- [ ] In the settings window in the model list, show the name of the repository the user chose. Using the repository url as key.
- [ ] In the model selection list in the main window, show the same repository name

Possible improvements after this:
- Allow users to "favourite" or tag models so they always appear at the top of the list.


jelmervdl:
Small update on where this is at: it works, but it crashes on a bad memory access. Not sure yet what is happening exactly, it might be something inside a signal handler that listens to the `externalRepositories` `valueChanged()` signal.

Steps to reproduce:

1. Start translateLocally
2. Go to translator settings
3. Go to the repository tab, add a new one via "import repository", doesn't matter what you type in
4. Click "OK" to save
5. EXC_BAD_ACCESS

```
Process 5248 stopped
* thread #1, queue = 'com.apple.main-thread', stop reason = EXC_BAD_ACCESS (code=1, address=0x353)
    frame #0: 0x0000000100053681 translateLocally`QtPrivate::QFunctorSlotObject<ModelManager::ModelManager(QObject*, Settings*)::$_0, 0, QtPrivate::List<>, void>::impl(int, QtPrivate::QSlotObjectBase*, QObject*, void**, bool*) [inlined] QArrayDataPointer<QList<QString> >::QArrayDataPointer(this=0x00007ff7bfefc7b0, other=0x0000000000000353) at qarraydatapointer.h:69:19 [opt]
   66  	    }
   67
   68  	    QArrayDataPointer(const QArrayDataPointer &other) noexcept
-> 69  	        : d(other.d), ptr(other.ptr), size(other.size)
   70  	    {
   71  	        ref();
   72  	    }
Target 0: (translateLocally) stopped.
(lldb) bt
* thread #1, queue = 'com.apple.main-thread', stop reason = EXC_BAD_ACCESS (code=1, address=0x353)
  * frame #0: 0x0000000100053681 translateLocally`QtPrivate::QFunctorSlotObject<ModelManager::ModelManager(QObject*, Settings*)::$_0, 0, QtPrivate::List<>, void>::impl(int, QtPrivate::QSlotObjectBase*, QObject*, void**, bool*) [inlined] QArrayDataPointer<QList<QString> >::QArrayDataPointer(this=0x00007ff7bfefc7b0, other=0x0000000000000353) at qarraydatapointer.h:69:19 [opt]
    frame #1: 0x0000000100053681 translateLocally`QtPrivate::QFunctorSlotObject<ModelManager::ModelManager(QObject*, Settings*)::$_0, 0, QtPrivate::List<>, void>::impl(int, QtPrivate::QSlotObjectBase*, QObject*, void**, bool*) [inlined] QArrayDataPointer<QList<QString> >::QArrayDataPointer(this=0x00007ff7bfefc7b0, other=0x0000000000000353) at qarraydatapointer.h:70:5 [opt]
    frame #2: 0x0000000100053681 translateLocally`QtPrivate::QFunctorSlotObject<ModelManager::ModelManager(QObject*, Settings*)::$_0, 0, QtPrivate::List<>, void>::impl(int, QtPrivate::QSlotObjectBase*, QObject*, void**, bool*) [inlined] QList<QList<QString> >::QList(this=0x00007ff7bfefc7b0, (null)=0x0000000000000353) at qlist.h:100:7 [opt]
    frame #3: 0x0000000100053681 translateLocally`QtPrivate::QFunctorSlotObject<ModelManager::ModelManager(QObject*, Settings*)::$_0, 0, QtPrivate::List<>, void>::impl(int, QtPrivate::QSlotObjectBase*, QObject*, void**, bool*) [inlined] QList<QList<QString> >::QList(this=0x00007ff7bfefc7b0, (null)=0x0000000000000353) at qlist.h:100:7 [opt]
    frame #4: 0x0000000100053681 translateLocally`QtPrivate::QFunctorSlotObject<ModelManager::ModelManager(QObject*, Settings*)::$_0, 0, QtPrivate::List<>, void>::impl(int, QtPrivate::QSlotObjectBase*, QObject*, void**, bool*) [inlined] SettingImpl<QList<QList<QString> > >::value(this=0x000000000000032b) const at Settings.h:70:16 [opt]
    frame #5: 0x0000000100053681 translateLocally`QtPrivate::QFunctorSlotObject<ModelManager::ModelManager(QObject*, Settings*)::$_0, 0, QtPrivate::List<>, void>::impl(int, QtPrivate::QSlotObjectBase*, QObject*, void**, bool*) [inlined] ModelManager::ModelManager(this=0x00006000002207d0)::$_0::operator()() const at ModelManager.cpp:90:52 [opt]
    frame #6: 0x0000000100053673 translateLocally`QtPrivate::QFunctorSlotObject<ModelManager::ModelManager(QObject*, Settings*)::$_0, 0, QtPrivate::List<>, void>::impl(int, QtPrivate::QSlotObjectBase*, QObject*, void**, bool*) [inlined] QtPrivate::FunctorCall<QtPrivate::IndexesList<>, QtPrivate::List<>, void, ModelManager::ModelManager(QObject*, Settings*)::$_0>::call(f=0x00006000002207d0)::$_0&, void**) at qobjectdefs_impl.h:146:13 [opt]
    frame #7: 0x0000000100053673 translateLocally`QtPrivate::QFunctorSlotObject<ModelManager::ModelManager(QObject*, Settings*)::$_0, 0, QtPrivate::List<>, void>::impl(int, QtPrivate::QSlotObjectBase*, QObject*, void**, bool*) [inlined] void QtPrivate::Functor<ModelManager::ModelManager(QObject*, Settings*)::$_0, 0>::call<QtPrivate::List<>, void>(f=0x00006000002207d0)::$_0&, void*, void**) at qobjectdefs_impl.h:256:13 [opt]
    frame #8: 0x0000000100053673 translateLocally`QtPrivate::QFunctorSlotObject<ModelManager::ModelManager(QObject*, Settings*)::$_0, 0, QtPrivate::List<>, void>::impl(which=<unavailable>, this_=0x00006000002207c0, r=0x00007ff7bfeff3b0, a=<unavailable>, ret=0x0000000000000000) at qobjectdefs_impl.h:420:17 [opt]
    frame #9: 0x0000000103363a43 QtCore`___lldb_unnamed_symbol1916$$QtCore + 803
    frame #10: 0x000000010000ba38 translateLocally`Setting::valueChanged(this=<unavailable>, _t1=<unavailable>, _t2=<unavailable>) at moc_Settings.cpp:136:5 [opt]
    frame #11: 0x000000010005be45 translateLocally`Setting::emitValueChanged(this=0x00007ff7bfeff3b0, name=<unavailable>, value=<unavailable>) at Settings.cpp:5:10 [opt]
    frame #12: 0x000000010006471f translateLocally`SettingImpl<QList<QList<QString> > >::setValue(this=0x00007ff7bfeff3b0, value=QList<QList<QString> > @ 0x00007ff7bfefca50, behavior=<unavailable>) at Settings.h:80:13 [opt]
    frame #13: 0x0000000100063f9f translateLocally`TranslatorSettingsDialog::applySettings(this=<unavailable>) at TranslatorSettingsDialog.cpp:93:30 [opt]
    frame #14: 0x0000000103363a43 QtCore`___lldb_unnamed_symbol1916$$QtCore + 803
    frame #15: 0x0000000102425992 QtWidgets`QDialog::done(int) + 82
    frame #16: 0x0000000103363a43 QtCore`___lldb_unnamed_symbol1916$$QtCore + 803
    frame #17: 0x00000001023b2311 QtWidgets`___lldb_unnamed_symbol3661$$QtWidgets + 209
    frame #18: 0x0000000103363d32 QtCore`___lldb_unnamed_symbol1916$$QtCore + 1554
    frame #19: 0x00000001022f5ecf QtWidgets`___lldb_unnamed_symbol1800$$QtWidgets + 111
    frame #20: 0x00000001022f5d6c QtWidgets`___lldb_unnamed_symbol1798$$QtWidgets + 236
    frame #21: 0x00000001022f6f59 QtWidgets`QAbstractButton::mouseReleaseEvent(QMouseEvent*) + 185
    frame #22: 0x00000001022170b6 QtWidgets`QWidget::event(QEvent*) + 390
    frame #23: 0x00000001021bf177 QtWidgets`QApplicationPrivate::notify_helper(QObject*, QEvent*) + 247
    frame #24: 0x00000001021c18c6 QtWidgets`QApplication::notify(QObject*, QEvent*) + 6086
    frame #25: 0x00000001033118f9 QtCore`QCoreApplication::notifyInternal2(QObject*, QEvent*) + 169
    frame #26: 0x00000001021bf93d QtWidgets`QApplicationPrivate::sendMouseEvent(QWidget*, QMouseEvent*, QWidget*, QWidget*, QWidget**, QPointer<QWidget>&, bool, bool) + 941
    frame #27: 0x000000010222fc2e QtWidgets`___lldb_unnamed_symbol789$$QtWidgets + 1662
    frame #28: 0x000000010222e98b QtWidgets`___lldb_unnamed_symbol784$$QtWidgets + 251
    frame #29: 0x00000001021bf177 QtWidgets`QApplicationPrivate::notify_helper(QObject*, QEvent*) + 247
    frame #30: 0x00000001021c02f5 QtWidgets`QApplication::notify(QObject*, QEvent*) + 501
    frame #31: 0x00000001033118f9 QtCore`QCoreApplication::notifyInternal2(QObject*, QEvent*) + 169
    frame #32: 0x0000000101681b7a QtGui`QGuiApplicationPrivate::processMouseEvent(QWindowSystemInterfacePrivate::MouseEvent*) + 2426
    frame #33: 0x00000001016d01cb QtGui`QWindowSystemInterface::sendWindowSystemEvents(QFlags<QEventLoop::ProcessEventsFlag>) + 427
    frame #34: 0x000000010383226b libqcocoa.dylib`___lldb_unnamed_symbol284$$libqcocoa.dylib + 43
    frame #35: 0x00007ff80e81819b CoreFoundation`__CFRUNLOOP_IS_CALLING_OUT_TO_A_SOURCE0_PERFORM_FUNCTION__ + 17
    frame #36: 0x00007ff80e818103 CoreFoundation`__CFRunLoopDoSource0 + 180
    frame #37: 0x00007ff80e817e7d CoreFoundation`__CFRunLoopDoSources0 + 242
    frame #38: 0x00007ff80e816898 CoreFoundation`__CFRunLoopRun + 892
    frame #39: 0x00007ff80e815e5c CoreFoundation`CFRunLoopRunSpecific + 562
    frame #40: 0x00007ff8174bd5e6 HIToolbox`RunCurrentEventLoopInMode + 292
    frame #41: 0x00007ff8174bd213 HIToolbox`ReceiveNextEventCommon + 283
    frame #42: 0x00007ff8174bd0e5 HIToolbox`_BlockUntilNextEventMatchingListInModeWithFilter + 70
    frame #43: 0x00007ff8112551fd AppKit`_DPSNextEvent + 927
    frame #44: 0x00007ff8112538ba AppKit`-[NSApplication(NSEvent) _nextEventMatchingEventMask:untilDate:inMode:dequeue:] + 1394
    frame #45: 0x00007ff811245f69 AppKit`-[NSApplication run] + 586
    frame #46: 0x0000000103831008 libqcocoa.dylib`___lldb_unnamed_symbol265$$libqcocoa.dylib + 2120
    frame #47: 0x000000010331abe6 QtCore`QEventLoop::exec(QFlags<QEventLoop::ProcessEventsFlag>) + 470
    frame #48: 0x0000000103311f22 QtCore`QCoreApplication::exec() + 130
    frame #49: 0x000000010000e529 translateLocally`main(argc=1, argv=0x00007ff7bfeff6a0) at main.cpp:58:12 [opt]
    frame #50: 0x0000000100e7d51e dyld`start + 462
```

XapaJIaMnu:
I can confirm the issue. Also seems PCRE2 build has changed so that also needs to be figured out.

XapaJIaMnu:
Also, something to do about existing installations on first run?
```
Could not parse model meta file "/home/dheart/.config/translateLocally/enes.student.tiny11-1645387947/.modelMeta.json" : file cannot be opened for reading
Could not parse model meta file "/home/dheart/.config/translateLocally/enbg.student.tiny11-1645371395/.modelMeta.json" : file cannot be opened for reading
Could not parse model meta file "/home/dheart/.config/translateLocally/bgen.student.tiny11-1645371287/.modelMeta.json" : file cannot be opened for reading
Could not parse model meta file "/home/dheart/.config/translateLocally/eng-fin.transformer-tiny11-jointvocab-1645622852/.modelMeta.json" : file cannot be opened for reading
Could not parse model meta file "/home/dheart/.config/translateLocally/ende.student.tiny11-1645387908/.modelMeta.json" : file cannot be opened for reading
Could not parse model meta file "/home/dheart/.config/translateLocally/esen.student.tiny11-1645387941/.modelMeta.json" : file cannot be opened for reading
Could not parse model meta file "/home/dheart/.config/translateLocally/encs.student.tiny11-1645458204/.modelMeta.json" : file cannot be opened for reading
Could not parse model meta file "/home/dheart/.config/translateLocally/deen.student.tiny11-1645387921/.modelMeta.json" : file cannot be opened for reading
Could not parse model meta file "/home/dheart/.config/translateLocally/eten.student.tiny11-1645622541/.modelMeta.json" : file cannot be opened for reading
```

XapaJIaMnu:
I've spent a bit of time looking into this. Beforehand, I believe the codepath taken was https://github.com/XapaJIaMnu/translateLocally/blob/4473caa3a2d48c755e021cc13ca07c69ffd98434/src/settings/Settings.h#L87
`AppendToValue`, and `RemoveFromValue` as opposed to `setValue`. 

However, even if I completely remove the code that modifies the `value_` the crash still happens when trying to emit the `QList<QStringList>`. Emiting the the new updated list, as opposed to the one saved in settings (`value`) also results in a crash.

XapaJIaMnu:
This is a workaround. Since `valueChanged` already carries it's necessary information.
```
diff --git a/src/inventory/ModelManager.cpp b/src/inventory/ModelManager.cpp
index 81fd009..542056f 100644
--- a/src/inventory/ModelManager.cpp
+++ b/src/inventory/ModelManager.cpp
@@ -86,8 +86,8 @@ ModelManager::ModelManager(QObject *parent, Settings * settings)
     
     // Update RepoManager with data from settings (also reload on update of setting)
     repositories_.load(settings->externalRepos.value());
-    connect(&(settings->externalRepos), &Setting::valueChanged, [&]{
-        repositories_.load(settings->externalRepos.value());
+    connect(&(settings->externalRepos), &Setting::valueChanged, this, [&](QString name, QVariant value){
+        repositories_.load(qvariant_cast<QList<QStringList>>(value));
         fetchRemoteModels(); // Are we sure we want to do this on update?
     });
```
Everything here screams "Race condition" to me. We update the QSetting backing and we immediately read out of it, however `QSetting` documentation says it is re-entrant and it shouldn't really have half committed....

jelmervdl:
Oh shit yes the settings pointer gets destroyed and then that reference no longer works! Thank you! You might also solve this by using [=] as capture I suppose?

XapaJIaMnu:
Yeah capturing by value should also work, HOWEVER, i find it a bit confusing to have a constructor to take `Settings` and not make it a member. I will leave it up to you to decide how to solve it, I don't feel too strongly about it. Also, since I read how `QSettings` backing works, I find that our `SettingImpl` is over-complicated. Writing a new Value to a `QSettings` field always overrides the previous, you can't append to it or remove elements from it efficiently. As such, I find that having a member `T value_` is redundant since its only purpose is to go to the backing value. Thus I have simplified it to:

```c++
/**
 * SettingImpl<T>:
 * Class that represents an individual setting. Similar to QProperty in Qt6.
 */
template <typename T>
class SettingImpl : public Setting {
private:
    QString name_;
    QSettings& backing_;

public:
    SettingImpl(QSettings &backing, QString name, T defaultValue = T())
    : name_(name),
      backing_(backing){}

    T value() const {
        return qvariant_cast<T>(backing_.value(name_));
    }

    void setValue(T newvalue, Behavior behavior = EmitWhenChanged) {
        if (behavior == EmitWhenChanged && newvalue == value())
            return;

        backing_.setValue(name_, QVariant::fromValue<T>(newvalue));
        emitValueChanged(name_, backing_.value(name_));
    }

    // Alias for value() to make Settings make look more like a normal Qt class.
    T operator()() const {
        return value();
    }
};
```

Here is the diff:
```patch
diff --git a/src/settings/Settings.h b/src/settings/Settings.h
index 9c68cbe..890c152 100644
--- a/src/settings/Settings.h
+++ b/src/settings/Settings.h
@@ -48,58 +48,23 @@ template <typename T>
 class SettingImpl : public Setting {
 private:
     QString name_;
-    T value_;
+    QSettings& backing_;
 
 public:
     SettingImpl(QSettings &backing, QString name, T defaultValue = T())
-    : name_(name) {
-        if constexpr (std::is_same_v<QList<QStringList>, T>) {
-            value_ = backing.value(name, QVariant::fromValue<QList<QStringList>>(QList<QStringList>())).value<QList<QStringList>>();
-        } else {
-            value_ = backing.value(name, defaultValue).template value<T>();
-        }
-        // When the value changes, also store it in QSettings
-        connect(this, &Setting::valueChanged, &backing, &QSettings::setValue);
-    }
+    : name_(name),
+      backing_(backing){}
 
     T value() const {
-        return value_;
-    }
-
-    void setValue(T value, Behavior behavior = EmitWhenChanged) {
-        if  constexpr (std::is_same_v<QList<QStringList>, T>) {
-            // Don't emit when the value doesn't change
-            if (behavior == EmitWhenChanged && value == value_)
-                return;
-
-            value_ = value;
-            emitValueChanged(name_, QVariant::fromValue<QList<QStringList>>(value_));
-        } else {
-            // Don't emit when the value doesn't change
-            if (behavior == EmitWhenChanged && value == value_)
-                return;
-
-            value_ = value;
-            emitValueChanged(name_, value_);
-        }
+        return qvariant_cast<T>(backing_.value(name_));
     }
 
-    void appendToValue(QStringList entry, Behavior behavior = EmitWhenChanged) {
-        if  constexpr (std::is_same_v<QList<QStringList>, T>) {
-            value_.append(entry);
-            emitValueChanged(name_, QVariant::fromValue<QList<QStringList>>(value_));
-        } else {
-            Q_ASSERT_X(true, "appendToValue", (std::string("Developer error: Attempted to append a setting entry to a non-appendable type: ") + std::string(typeid(T).name())).c_str());
-        }
-    }
+    void setValue(T newvalue, Behavior behavior = EmitWhenChanged) {
+        if (behavior == EmitWhenChanged && newvalue == value())
+            return;
 
-    void removeFromValue(int index, Behavior behavior = EmitWhenChanged) {
-        if  constexpr (std::is_same_v<QList<QStringList>, T>) {
-            value_.removeAt(index);
-            emitValueChanged(name_, QVariant::fromValue<QList<QStringList>>(value_));
-        } else {
-            Q_ASSERT_X(true, "removeFromValue", (std::string("Developer error: Attempted to remove a setting entry to a non-removable type") + std::string(typeid(T).name())).c_str());
-        }
+        backing_.setValue(name_, QVariant::fromValue<T>(newvalue));
+        emitValueChanged(name_, backing_.value(name_));
     }
 
     // Alias for value() to make Settings make look more like a normal Qt class.
```
The main behavioural difference is that writing to the `QSettings backend_` now happens in the main thread, as opposed to delegating it to a signal, but it should be instantaneous, and is more intuitive?

XapaJIaMnu:
So, is this ready for review?

jelmervdl:
I think it is ready for review now. But please do poke it quite a bit.

src/inventory/ModelManager.cpp
```diff
@@ -394,6 +422,50 @@ void ModelManager::scanForModels(QString path) {
     updateAvailableModels();
 }
 
+bool ModelManager::readModelMetaFromDir(ModelMeta &model, QString dir) const {
+    QFile metaFile(dir + "/.modelMeta.json");
+    if (!metaFile.open(QIODevice::ReadOnly | QIODevice::Text)) {
+        qDebug() << "Could not parse model meta file" << metaFile.fileName() << ": file cannot be opened for reading";
```
XapaJIaMnu, line 428: We should remove that, it spits a lot of noise:
`Could not parse model meta file "~/.config/translateLocally/bgen.student.tiny11-1656082614/.modelMeta.json" : file cannot be opened for reading`
If modelMeta is not there, we should generate it to the best of our ability.

src/mainwindow.cpp
```diff
@@ -320,7 +321,15 @@ void MainWindow::downloadModel(Model model) {
     ui_->cancelDownloadButton->setEnabled(true);
     showDownloadPane(true);
 
-    QNetworkReply *reply = network_.downloadFile(model.url, QCryptographicHash::Sha256, model.checksum);
+    // Pass on meta info about the model so we remember once we've downloaded it
+    ModelMeta meta;
+    meta.modelUrl = model.url;
+    meta.repositoryUrl = model.repositoryUrl;
+
+    qDebug() << "Downloading:";
```
XapaJIaMnu, line 329: This should probably be removed, we don't want it:
`shortName: "en-bg-tiny" modelName: "English-Bulgarian tiny" url: "https://data.statmt.org/bergamot/models/bgen/enbg.student.tiny11.tar.gz" path: "" src: "English" trg: "Bulgarian" type: "tiny" localversion -1 localAPI -1 remoteversion: 1 remoteAPI 1
`

src/inventory/ModelManager.cpp
```diff
@@ -82,9 +84,24 @@ ModelManager::ModelManager(QObject *parent, Settings * settings)
             QDir().mkpath(configDir_.absolutePath());
         }
     }
+    
+    // Update RepoManager with data from settings (also reload on update of setting)
+    repositories_.load(settings_->externalRepos.value());
```
XapaJIaMnu, line 89: We should probably notify the user that once a new repo is added, they need to explicitly click "Download new models".

src/inventory/ModelManager.cpp
```diff
@@ -372,17 +395,22 @@ void ModelManager::scanForModels(QString path) {
...
-            } else {
-                // We have a folder in our models directory that doesn't contain a model. This is ok.
+            
+            // We have a folder in our models directory that doesn't contain a model. This is ok.
+            if (obj.empty())
+                continue;
+            
+            auto model = parseModelInfo(obj);
+            if (!model) {
+                emit error(tr("Corrupted json file: %1/model_info.json. Delete or redownload.").arg(current));
```
XapaJIaMnu, line 405: Should we make this a messagebox where ok deletes the model? Or too much plumbing. This scenario would be quite unlikely I feel

src/inventory/ModelManager.cpp
```diff
@@ -394,6 +422,50 @@ void ModelManager::scanForModels(QString path) {
...
+bool ModelManager::writeModelMetaToDir(ModelMeta const &model, QString dir) const {
+    QJsonDocument json{QJsonObject{
+        {"modelUrl", model.modelUrl},
+        {"repositoryUrl", model.repositoryUrl},
+        {"installedOn", model.installedOn.toString(Qt::ISODate)},
+    }};
+
+    QFile metaFile(dir + "/.modelMeta.json");
+    if (!metaFile.open(QIODevice::WriteOnly | QIODevice::Text)) {
+        qDebug() << "Could not write model meta file" << metaFile.fileName() << ": file cannot be opened for writing";
```
XapaJIaMnu, line 459: qDebug() -> error()?

src/inventory/ModelManager.h
```diff
@@ -24,17 +24,28 @@ namespace translateLocally {
     }
 }
 
-struct Model {
+/**
+ * @Brief outside info about a model that it cannot about itself in a shipped
```
XapaJIaMnu, line 28: `it cannot` -> `it cannot know`?

src/inventory/ModelManager.h
```diff
@@ -78,24 +85,24 @@ struct Model {
...
             if (key == "srcTags") {
                 srcTags = val.toVariantMap();
             } else {
-               parseError = true;
+               parseError = true; // TODO Again
             }
         }
         if (parseError) {
-            std::cerr << "Unknown key type. " << key.toStdString() << " Something is very wrong!" << std::endl;
+            qDebug() << "Model::set() was called with an unknown key type for key" << key;
```
XapaJIaMnu, line 98: we shouldn't be spitting on the command line, right? Although this is probably a developer error, if we end here?

src/inventory/RepoManager.cpp
```diff
@@ -1,118 +1,36 @@
...
-
-    for (auto &&index : rows) {
-        if (!canRemove(index)) {
-            emit warning("Unable to remove the builtin repository.");
+    repositories_.insert(kDefaultRepositoryURL, {kDefaultRepositoryName,kDefaultRepositoryURL, true});
+    
+    for (auto &&pair : data) {
+        // Future proofing: make sure we don't load repositories that already exist
+        // as default repositories.
+        if (repositories_.contains(pair.back()))
```
XapaJIaMnu, line 20: Note to self, see how this works exactly

src/settings/RepositoryTableModel.cpp
```diff
@@ -0,0 +1,151 @@
...
+        // pair: [name, ..., url] where ... is currently [].
+
+        if (urls_.contains(pair.back()))
+            continue;
+
+        repositories_ << Repository{pair.first(), pair.back(), false};
+        urls_.insert(pair.back());
+    }
+        
+    // TODO: I'm lying here, it has already happened. Will Qt be okay with that?
```
XapaJIaMnu, line 37: QT will start having trust issues at this point, if you ask me...

src/settings/RepositoryTableModel.cpp
```diff
@@ -0,0 +1,151 @@
...
+    endInsertRows();
+}
+
+bool RepositoryTableModel::canRemove(QModelIndex index) const {
+    return !repositories_.at(index.row()).isDefault;
+}
+
+void RepositoryTableModel::insert(QString name, QString url) {
+    if (urls_.contains(url)) {
+        // TODO: update name instead?
```
XapaJIaMnu, line 48: The user can just delete and add again, too much work otherwise.

XapaJIaMnu (commented):
First round review. It will need more poking, which i will do later today.

src/inventory/ModelManager.cpp
jelmervdl: I don't think there's much value of generating it over just skipping reading it. Trying to generate it could have a whole bunch of error cases on its own, such as read only filesystem, permission issues, etc.

How about instead of removing debug messages, we keep them but hide them behind a `--debug` flag?

src/inventory/ModelManager.cpp
jelmervdl: There is already validation on download, so this could be a corruption scenario. But more likely it is someone trying to make a model that works with translateLocally before uploading it somewhere. With that in mind, I'm adding a better description of what might be wrong with the file.

src/inventory/ModelManager.h
jelmervdl: Yeah, developer error but again one that doesn't do any harm. We have `validateModel()` already for checking whether a model has all necessary data. This one is only about calling `set()` with the wrong type, or with an unknown `key`.

And again, we can easily hide this behind a `--debug` flag

src/settings/RepositoryTableModel.cpp
jelmervdl: I agree. Maybe at some point I'll make the table properly editable so you can double-click a field and change the names or urls that way.

src/settings/RepositoryTableModel.cpp
```diff
@@ -0,0 +1,151 @@
...
+QList<QStringList> RepositoryTableModel::dump() const {
+    QList<QStringList> out;
+    for (auto &&repo : repositories_) {
+        if (repo.isDefault) continue;
+        out << QStringList{repo.name, repo.url};
+    }
+    return out;
+}
+
+void RepositoryTableModel::load(QList<QStringList> data) {
```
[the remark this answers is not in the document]
jelmervdl: Somewhat, but that's the (small) price to pay for storing it as a StringList in the settings, and making Settings the single source of truth for now. This also allows the internal data model for RepositoryTableManager and RepoManager to be different, which makes those things a little more self contained.

src/inventory/ModelManager.cpp
jelmervdl: It's implicitly communicated right now. The code just below that bit clears `remoteModels_`, the list will only show local ones and the _Download new models_ option after changing repositories:
```
        // We do clear the remoteModels list so that it is clear that it is
        // outdated and/or incomplete. Now users can click the "download model
        // list" again and make an informed decision to access the internet.
        remoteModels_.clear();
        updateAvailableModels();
```
A popup saying "hey click that button" every time repositories are updated is probably annoying. And/or going to be ignored. Nobody ever reads popups…

jelmervdl: Or to be more clear about it: they have _no other option_ than to click that "download new models" option if they want to download any model at all after updating the repositories. The flow is exactly like they're already used to at that point.

XapaJIaMnu: Ok, fair point.

src/settings/RepositoryTableModel.cpp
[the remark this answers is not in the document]
XapaJIaMnu: Oook, I'm getting a bit confused here. So. When loading the model, `RepoManager` inserts the default repository. All other repository insertions are managed by `RepoTableModel`?
`RepoManager` just serves as a container ObjectEssentially to feed `Repositories` inside `ModelManager`. Honestly, we can just remove most of it and rely on `RepositoryTableModel`.  What do you think?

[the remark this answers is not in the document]
XapaJIaMnu: I did some playing around:
ModelManager and Mainwindow can both access `settings_->externalRepos()`. If we make the default contain the default repository (eg settings_->externalRepos() becomes settings_->repos()) and prevent it from being deleteable and make everything read it directly out of settings, we don't need `inventory/RepoManager*`. This is a small refactoring and I'm happy to keep it out of this PR, so t we don't block anything else and do it later in separate PR. If you agree with me, I'm happy to have this merged now.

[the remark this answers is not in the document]
jelmervdl: You're right, RepoManager is just a substructure of ModelManager at this point. RepoManager exists to add that default entry. I suppose that is something Settings can do very well on its own as well. I'll change that, it makes things simpler :)

One of the reasons for the current set-up, in which RepositoryTableModel is not the single source of truth, is to allow you to edit the table in the settings panel, but then press "Cancel" and not have those changes take effect. It forces Settings to be the source of truth, and allows us to control when that one is actually updated[^1].

[^1]: The translation settings window is a bit of an inconsistent thing right now anyway: Changes in the model table do have immediate effect, but all the other panes in that window do not.

I'm also not really a fan of how the ModelManager interface looks right now. It does too many things… and I wouldn't mind if I can move the bits that provide the data for rendering the table and the dropdown menu to their own classes. I don't know how I could make Settings also the single source here, since adding/removing models is much more involved than updating a value in a config file.

XapaJIaMnu:
Suggested improvement, can go in a different pr
`settings.externalRepos` -> `repos` with our default repositories part of the default constructors.
`settings.externalRepos` `QList<QStringList>` -> `QMap<QString, Repository>` with proper  plumbing so that we can serialise it. (Not sure if this is worth it it tbh, but it doesn't look like that much extra code: https://stackoverflow.com/questions/18144377/writing-and-reading-custom-class-to-qsettings )
That will allow us to remove the temporary `repositories_` inside modelManager.

jelmervdl:
> I say merge this for now, merge the other open PR and then make small incremental changes.

Gonna go with this 👍

PR #103, opened 2022-05-04
Add configure command to native messaging interface

jelmervdl:
Useful for benchmarking mostly

### DIFF
--- a/src/cli/NativeMsgIface.cpp
+++ b/src/cli/NativeMsgIface.cpp
@@ -224,6 +224,17 @@ void NativeMsgIface::handleRequest(ListRequest request)  {
     writeResponse(request, modelsJson);
 }
 
+void NativeMsgIface::handleRequest(ConfigureRequest request)  {
+    marian::bergamot::AsyncService::Config serviceConfig;
+    serviceConfig.numWorkers = request.threads;
+    serviceConfig.cacheSize = request.cacheSize;
+    service_.reset();
+    service_ = std::make_shared<marian::bergamot::AsyncService>(serviceConfig);
+    
+    QJsonObject response{}; // I don't know...
+    writeResponse(request, response);
+}
+
 void NativeMsgIface::handleRequest(DownloadRequest request)  {
     // Edge case: client issued a DownloadRequest before fetching the list of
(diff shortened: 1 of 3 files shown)


src/cli/NativeMsgIface.cpp
```diff
@@ -359,13 +370,21 @@ request_variant NativeMsgIface::parseJsonInput(QByteArray input) {
         ret.id = id;
         for (auto&& key : mandatoryKeysDownload) {
             QJsonValueRef val = data[key];
-            if (val.isNull()) {
+            if (val.isNull() || val.isUndefined()) {
```
XapaJIaMnu, line 373: So... `isNull()` is like `key: ""``, whereas `isUndefined()` is the key is not found at all?

jelmervdl: > The returned [QJsonValue](https://doc.qt.io/qt-6/qjsonvalue.html) is [QJsonValue::Undefined](https://doc.qt.io/qt-6/qjsonvalue.html#Type-enum) if the key does not exist.

https://doc.qt.io/qt-6/qjsonobject.html#value

XapaJIaMnu: How was it then working beforehand!?

jelmervdl:
I marked it as draft because I just implemented this as a one-off for some benchmarking work. However, this functionality might be useful for other use cases as well.

There should be probably be a `min(requested-threads, std::thread::hardware_concurrency())` or something.

XapaJIaMnu:
Yeah, that's what I do: https://github.com/XapaJIaMnu/translateLocally/blob/master/src/settings/TranslatorSettingsDialog.cpp#L24
Should that be written to settings if changed?

jelmervdl:
Oeh note to self: it should do everything from the pointer reset line in std::async or something, because that's a blocking call that will wait for all still pending translations to finish. And that should not be blocking the event loop!

And then I need to think a bit about how it should treat translation requests that come in after the Configure command. They should be queued/processed after it, and not to the service that is about to be reset (thereby delaying the reset even more). So maybe the reset() should be blocking the event loop?

PR #112, opened 2022-07-11
Make settings.repos the single source of truth

XapaJIaMnu:
So, what i did here is I made `settings_.repos` the single source of truth. Also removed all intermittent `QMap<QString, translateLocally::Repository`. I like PRs that remove more code than they add, ~~but sadly `removeRows` acts weirdly and I am not sure why. It always removes less rows than it has to and mistakenly runs into the default repository.~~

Also, due to using QMap, ordering is.. Random (Sorted alphabetically by the key, which is the URL). Not sure how much we care about this case, as it just affects the way repositories are displayed in the settings.


jelmervdl:
I'm having issues with this pull request:
```
translateLocally(24375,0x100ef8600) malloc: Heap corruption detected, free list is damaged at 0x6000021443c0
*** Incorrect guard value: 209653941748674
translateLocally(24375,0x100ef8600) malloc: *** set a breakpoint in malloc_error_break to debug
```
I'm trying to debug it now, but the corruption does not seem to be related to the place that triggers the detection Both clicking the "Download more…" in the models dropdown in the main window, and opening the "Languages" tab in the settings, trigger it.

Trace from triggering it through the main window dropdown:
```
Process 24925 stopped
* thread #1, queue = 'com.apple.main-thread', stop reason = signal SIGABRT
    frame #0: 0x00007ff8169e900e libsystem_kernel.dylib`__pthread_kill + 10
libsystem_kernel.dylib`__pthread_kill:
->  0x7ff8169e900e <+10>: jae    0x7ff8169e9018            ; <+20>
    0x7ff8169e9010 <+12>: movq   %rax, %rdi
    0x7ff8169e9013 <+15>: jmp    0x7ff8169e31c5            ; cerror_nocancel
    0x7ff8169e9018 <+20>: retq
Target 0: (translateLocally) stopped.
(lldb) bt
* thread #1, queue = 'com.apple.main-thread', stop reason = signal SIGABRT
  * frame #0: 0x00007ff8169e900e libsystem_kernel.dylib`__pthread_kill + 10
    frame #1: 0x00007ff816a1f1ff libsystem_pthread.dylib`pthread_kill + 263
    frame #2: 0x00007ff81696ad24 libsystem_c.dylib`abort + 123
    frame #3: 0x00007ff816848357 libsystem_malloc.dylib`malloc_vreport + 551
    frame #4: 0x00007ff81685c308 libsystem_malloc.dylib`malloc_zone_error + 178
    frame #5: 0x00007ff81683b0e8 libsystem_malloc.dylib`nanov2_allocate_from_block + 582
    frame #6: 0x00007ff81683a677 libsystem_malloc.dylib`nanov2_allocate + 130
    frame #7: 0x00007ff81683a5a4 libsystem_malloc.dylib`nanov2_malloc + 56
    frame #8: 0x00007ff816856abb libsystem_malloc.dylib`_malloc_zone_malloc + 125
    frame #9: 0x0000000103427dfe QtCore`QArrayData::allocate(QArrayData**, long long, long long, long long, QArrayData::AllocationOption) + 222
    frame #10: 0x00000001033f5a89 QtCore`QString::QString(long long, Qt::Initialization) + 73
    frame #11: 0x0000000103405121 QtCore`___lldb_unnamed_symbol2658$$QtCore + 145
    frame #12: 0x0000000103404d62 QtCore`QString::arg(QStringView, int, QChar) const + 130
    frame #13: 0x0000000103404cca QtCore`QString::arg(QString const&, int, QChar) const + 42
    frame #14: 0x0000000100018ed5 translateLocally`MainWindow::updateLocalModels(this=0x00007ff7bfefeef8) at mainwindow.cpp:395:48 [opt]
    frame #15: 0x0000000103363a43 QtCore`___lldb_unnamed_symbol1916$$QtCore + 803
    frame #16: 0x0000000100053f70 translateLocally`ModelManager::updateAvailableModels(this=0x00007ff7bfeff398) at ModelManager.cpp:719:10 [opt]
    frame #17: 0x0000000100056e1c translateLocally`ModelManager::parseRemoteModels(this=0x00007ff7bfeff398, obj=QJsonObject @ 0x00007ff7bfefd010, repositoryUrl=QString @ 0x00007ff7bfefcff8) at ModelManager.cpp:659:5 [opt]
    frame #18: 0x000000010005a825 translateLocally`QtPrivate::QFunctorSlotObject<ModelManager::fetchRemoteModels(QVariant)::$_3, 0, QtPrivate::List<>, void>::impl(int, QtPrivate::QSlotObjectBase*, QObject*, void**, bool*) at ModelManager.cpp:615:21 [opt]
    frame #19: 0x000000010005a4f8 translateLocally`QtPrivate::QFunctorSlotObject<ModelManager::fetchRemoteModels(QVariant)::$_3, 0, QtPrivate::List<>, void>::impl(int, QtPrivate::QSlotObjectBase*, QObject*, void**, bool*) [inlined] QtPrivate::FunctorCall<QtPrivate::IndexesList<>, QtPrivate::List<>, void, ModelManager::fetchRemoteModels(QVariant)::$_3>::call(f=0x0000600002950780)::$_3&, void**) at qobjectdefs_impl.h:146:13 [opt]
    frame #20: 0x000000010005a4f8 translateLocally`QtPrivate::QFunctorSlotObject<ModelManager::fetchRemoteModels(QVariant)::$_3, 0, QtPrivate::List<>, void>::impl(int, QtPrivate::QSlotObjectBase*, QObject*, void**, bool*) [inlined] void QtPrivate::Functor<ModelManager::fetchRemoteModels(QVariant)::$_3, 0>::call<QtPrivate::List<>, void>(f=0x0000600002950780)::$_3&, void*, void**) at qobjectdefs_impl.h:256:13 [opt]
    frame #21: 0x000000010005a4f8 translateLocally`QtPrivate::QFunctorSlotObject<ModelManager::fetchRemoteModels(QVariant)::$_3, 0, QtPrivate::List<>, void>::impl(which=<unavailable>, this_=0x0000600002950770, r=<unavailable>, a=<unavailable>, ret=<unavailable>) at qobjectdefs_impl.h:420:17 [opt]
    frame #22: 0x0000000103363a43 QtCore`___lldb_unnamed_symbol1916$$QtCore + 803
    frame #23: 0x000000010152b12d QtNetwork`___lldb_unnamed_symbol2156$$QtNetwork + 589
    frame #24: 0x000000010335adf9 QtCore`QObject::event(QEvent*) + 873
    frame #25: 0x00000001021bf177 QtWidgets`QApplicationPrivate::notify_helper(QObject*, QEvent*) + 247
    frame #26: 0x00000001021c02f5 QtWidgets`QApplication::notify(QObject*, QEvent*) + 501
    frame #27: 0x00000001033118f9 QtCore`QCoreApplication::notifyInternal2(QObject*, QEvent*) + 169
    frame #28: 0x0000000103312a9e QtCore`QCoreApplicationPrivate::sendPostedEvents(QObject*, int, QThreadData*) + 814
    frame #29: 0x0000000103016af5 libqcocoa.dylib`___lldb_unnamed_symbol272$$libqcocoa.dylib + 325
    frame #30: 0x000000010301726b libqcocoa.dylib`___lldb_unnamed_symbol284$$libqcocoa.dylib + 43
    frame #31: 0x00007ff816ae61ab CoreFoundation`__CFRUNLOOP_IS_CALLING_OUT_TO_A_SOURCE0_PERFORM_FUNCTION__ + 17
    frame #32: 0x00007ff816ae6113 CoreFoundation`__CFRunLoopDoSource0 + 180
    frame #33: 0x00007ff816ae5e8d CoreFoundation`__CFRunLoopDoSources0 + 242
    frame #34: 0x00007ff816ae48a8 CoreFoundation`__CFRunLoopRun + 892
    frame #35: 0x00007ff816ae3e6c CoreFoundation`CFRunLoopRunSpecific + 562
    frame #36: 0x00007ff81f7925e6 HIToolbox`RunCurrentEventLoopInMode + 292
    frame #37: 0x00007ff81f79234a HIToolbox`ReceiveNextEventCommon + 594
    frame #38: 0x00007ff81f7920e5 HIToolbox`_BlockUntilNextEventMatchingListInModeWithFilter + 70
    frame #39: 0x00007ff81951dfad AppKit`_DPSNextEvent + 927
    frame #40: 0x00007ff81951c66a AppKit`-[NSApplication(NSEvent) _nextEventMatchingEventMask:untilDate:inMode:dequeue:] + 1394
    frame #41: 0x00007ff81950ed19 AppKit`-[NSApplication run] + 586
    frame #42: 0x0000000103016008 libqcocoa.dylib`___lldb_unnamed_symbol265$$libqcocoa.dylib + 2120
    frame #43: 0x000000010331abe6 QtCore`QEventLoop::exec(QFlags<QEventLoop::ProcessEventsFlag>) + 470
    frame #44: 0x0000000103311f22 QtCore`QCoreApplication::exec() + 130
    frame #45: 0x000000010000e282 translateLocally`main(argc=2, argv=0x00007ff7bfeff640) at main.cpp:64:12 [opt]
    frame #46: 0x0000000100e7d52e dyld`start + 462
```

Trace from triggering it through the settings window:
```
translateLocally(24932,0x100ef8600) malloc: Heap corruption detected, free list is damaged at 0x600002142350
*** Incorrect guard value: 209653941740370
translateLocally(24932,0x100ef8600) malloc: *** set a breakpoint in malloc_error_break to debug
Process 24932 stopped
* thread #1, queue = 'com.apple.main-thread', stop reason = signal SIGABRT
    frame #0: 0x00007ff8169e900e libsystem_kernel.dylib`__pthread_kill + 10
libsystem_kernel.dylib`__pthread_kill:
->  0x7ff8169e900e <+10>: jae    0x7ff8169e9018            ; <+20>
    0x7ff8169e9010 <+12>: movq   %rax, %rdi
    0x7ff8169e9013 <+15>: jmp    0x7ff8169e31c5            ; cerror_nocancel
    0x7ff8169e9018 <+20>: retq
Target 0: (translateLocally) stopped.
(lldb) bt
* thread #1, queue = 'com.apple.main-thread', stop reason = signal SIGABRT
  * frame #0: 0x00007ff8169e900e libsystem_kernel.dylib`__pthread_kill + 10
    frame #1: 0x00007ff816a1f1ff libsystem_pthread.dylib`pthread_kill + 263
    frame #2: 0x00007ff81696ad24 libsystem_c.dylib`abort + 123
    frame #3: 0x00007ff816848357 libsystem_malloc.dylib`malloc_vreport + 551
    frame #4: 0x00007ff81685c308 libsystem_malloc.dylib`malloc_zone_error + 178
    frame #5: 0x00007ff81683b0e8 libsystem_malloc.dylib`nanov2_allocate_from_block + 582
    frame #6: 0x00007ff81683a677 libsystem_malloc.dylib`nanov2_allocate + 130
    frame #7: 0x00007ff81683c68f libsystem_malloc.dylib`nanov2_calloc + 126
    frame #8: 0x00007ff816856b75 libsystem_malloc.dylib`_malloc_zone_calloc + 60
    frame #9: 0x00007ff81bda6f9c CoreGraphics`CGGStateRenderingReset + 59
    frame #10: 0x00007ff81bda6c69 CoreGraphics`CGGStateReset + 40
    frame #11: 0x00007ff81bda6b5d CoreGraphics`CGGStateCreate + 83
    frame #12: 0x00007ff81bda6aed CoreGraphics`CGGStackGetGState + 23
    frame #13: 0x00007ff81bda670e CoreGraphics`CGContextCreateWithDelegateAndInfo + 174
    frame #14: 0x00007ff81bda65d0 CoreGraphics`bitmap_context_create + 46
    frame #15: 0x00007ff81bda5c4a CoreGraphics`CGBitmapContextCreate + 29
    frame #16: 0x0000000101995525 QtGui`QMacCGContext::initialize(QImage const*, QPainter*) + 149
    frame #17: 0x00000001019957a2 QtGui`QMacCGContext::QMacCGContext(QPainter*) + 194
    frame #18: 0x0000000103f2a539 libqmacstyle.dylib`___lldb_unnamed_symbol90$$libqmacstyle.dylib + 105
    frame #19: 0x00000001024bc82c QtWidgets`QStyledItemDelegate::paint(QPainter*, QStyleOptionViewItem const&, QModelIndex const&) const + 124
    frame #20: 0x00000001024e6c83 QtWidgets`___lldb_unnamed_symbol7054$$QtWidgets + 355
    frame #21: 0x00000001024e88ef QtWidgets`QTableView::paintEvent(QPaintEvent*) + 1951
    frame #22: 0x0000000102217374 QtWidgets`QWidget::event(QEvent*) + 1092
    frame #23: 0x00000001022bc72b QtWidgets`QFrame::event(QEvent*) + 43
    frame #24: 0x000000010248a3a1 QtWidgets`QAbstractItemView::viewportEvent(QEvent*) + 1169
    frame #25: 0x0000000103311b98 QtCore`QCoreApplicationPrivate::sendThroughObjectEventFilters(QObject*, QEvent*) + 232
    frame #26: 0x00000001021bf163 QtWidgets`QApplicationPrivate::notify_helper(QObject*, QEvent*) + 227
    frame #27: 0x00000001021c02f5 QtWidgets`QApplication::notify(QObject*, QEvent*) + 501
    frame #28: 0x00000001033118f9 QtCore`QCoreApplication::notifyInternal2(QObject*, QEvent*) + 169
    frame #29: 0x000000010220929a QtWidgets`QWidgetPrivate::drawWidget(QPaintDevice*, QRegion const&, QPoint const&, QFlags<QWidgetPrivate::DrawWidgetFlag>, QPainter*, QWidgetRepaintManager*) + 3466
    frame #30: 0x00000001022291d4 QtWidgets`___lldb_unnamed_symbol733$$QtWidgets + 4516
    frame #31: 0x000000010222960f QtWidgets`___lldb_unnamed_symbol734$$QtWidgets + 271
    frame #32: 0x000000010221754a QtWidgets`QWidget::event(QEvent*) + 1562
    frame #33: 0x00000001021bf177 QtWidgets`QApplicationPrivate::notify_helper(QObject*, QEvent*) + 247
    frame #34: 0x00000001021c02f5 QtWidgets`QApplication::notify(QObject*, QEvent*) + 501
    frame #35: 0x00000001033118f9 QtCore`QCoreApplication::notifyInternal2(QObject*, QEvent*) + 169
    frame #36: 0x0000000103312a9e QtCore`QCoreApplicationPrivate::sendPostedEvents(QObject*, int, QThreadData*) + 814
    frame #37: 0x0000000103116af5 libqcocoa.dylib`___lldb_unnamed_symbol272$$libqcocoa.dylib + 325
    frame #38: 0x000000010311726b libqcocoa.dylib`___lldb_unnamed_symbol284$$libqcocoa.dylib + 43
    frame #39: 0x00007ff816ae61ab CoreFoundation`__CFRUNLOOP_IS_CALLING_OUT_TO_A_SOURCE0_PERFORM_FUNCTION__ + 17
    frame #40: 0x00007ff816ae6113 CoreFoundation`__CFRunLoopDoSource0 + 180
    frame #41: 0x00007ff816ae5e8d CoreFoundation`__CFRunLoopDoSources0 + 242
    frame #42: 0x00007ff816ae48a8 CoreFoundation`__CFRunLoopRun + 892
    frame #43: 0x00007ff816ae3e6c CoreFoundation`CFRunLoopRunSpecific + 562
    frame #44: 0x00007ff81f7925e6 HIToolbox`RunCurrentEventLoopInMode + 292
    frame #45: 0x00007ff81f792213 HIToolbox`ReceiveNextEventCommon + 283
    frame #46: 0x00007ff81f7920e5 HIToolbox`_BlockUntilNextEventMatchingListInModeWithFilter + 70
    frame #47: 0x00007ff81951dfad AppKit`_DPSNextEvent + 927
    frame #48: 0x00007ff81951c66a AppKit`-[NSApplication(NSEvent) _nextEventMatchingEventMask:untilDate:inMode:dequeue:] + 1394
    frame #49: 0x00007ff81950ed19 AppKit`-[NSApplication run] + 586
    frame #50: 0x0000000103116008 libqcocoa.dylib`___lldb_unnamed_symbol265$$libqcocoa.dylib + 2120
    frame #51: 0x000000010331abe6 QtCore`QEventLoop::exec(QFlags<QEventLoop::ProcessEventsFlag>) + 470
    frame #52: 0x0000000103311f22 QtCore`QCoreApplication::exec() + 130
    frame #53: 0x000000010000e282 translateLocally`main(argc=2, argv=0x00007ff7bfeff640) at main.cpp:64:12 [opt]
    frame #54: 0x0000000100e7d52e dyld`start + 462
```

XapaJIaMnu:
Hmm, I can't seem to trigger it on Linux. Could it be a mac only issue? Also, could it be you would need a clean settings/prefix. I removed some unnecessary structs that may still be in an old config file (Though on linux it just resulted in a harmless warning in the console)

jelmervdl:
It looks like it is caused by the `auto repo = models_.getRepository(model)` line. Even if I never use `repo` after that, it corrupts the heap somehow.

I think it is caused by `Settings::repos()` returning an rvalue, so in this bit of code, `find()` and `end()` work on different copies. Also, using `find()` this way looks illegal to me because isn't the object `it` points to gone after that `find` call?

```cpp
std::optional<Repository> ModelManager::getRepository(const Model &model) const {
    auto it = settings_->repos().find(model.repositoryUrl);
    if (it == settings_->repos().end())
        return std::nullopt;
    return *it;
}
```

Replacing it with something like this works:
```cpp
std::optional<Repository> ModelManager::getRepository(const Model &model) const {
    auto repos = settings_->repos();
    auto it = repos.find(model.repositoryUrl);
    if (it == repos.end())
        return std::nullopt;
    return *it;
}
```

But I'm also a bit worried that it just hides a symptom of a mayor inefficiency.

jelmervdl:
Maybe it is an implementation difference in QSettings between Qt5 and Qt6, or when using the macOS preferences system vs Linux' ini files I guess.

jelmervdl:
Ugh, what a mess. If we bring back the local internal copy in SettingImpl, and make SettingImpl::value() return a reference, we might just make the whole class thread-unsafe because there is no guarantee that the setting isn't changed while some other bit of code still assumes the reference is still valid.

I'm in favour of keeping the current implementation that copies. I would have liked my linter to highlight troublesome lifetime issues… but its C++ not rust so I guess I can keep dreaming for a while.

I'll go through all references to Settings::repos() and make sure that they use a single instance if they're doing things like `find()`.

jelmervdl:
Re speed: yes table is probably always small, but `ModelManager::getRepository()` is called an awful lot. E.g. every table row in the "Languages" tab, every menu item in the main window dropdown.

It would be good it we could optimise that call… but that would basically undo 50% of this pull request.

XapaJIaMnu:
Ugh... I really, really, really, dislike that.

We could offload it to use the ModelMeta.json? That way the lookup is done once during download?

jelmervdl:
> We could offload it to use the ModelMeta.json? That way the lookup is done once during download?

If you then later change the name or the url of the repository, that change would then not be reflected in the Languages table. Or partially, depending on when you downloaded models. And we would have this exact same inefficiency problem again if we want to attach other functionality to that model <-> repo relation in the future.

We seem to have 4 options:
1. Make local copies of the repos map for the `find()` + `end()` use case. Does not solve inefficiency of creating the Map on every `repos()` call.
2. Store a copy of `repos()` as a member variable of `ModelManager`. Solves inefficiency for this particular case, but gone is the whole single source of truth thing.
3. Store a copy of the value in the `SettingImpl` class. Solves inefficiency and we can return `const &`. `Settings` is still single source of truth. Problem is that in theory you don't know when that reference goes out of scope as the setting can be updated at any time. I don't think we have any code paths that could cause this right now. I'm also thinking that there must be a way to use signals/slots to sync all settings changes to the main/ui thread which would then guarantee that the `find()` + `end()` scenario would always work as expected. But as mentioned, not strictly necessary right now.
4. Store repository name in modelInfo.json or even the struct at load time. Downside is that any changes to the repos() setting are then not reflected in the dropdown or Languages table. That makes this pretty much a no for me.

My preference is 3 right now. It might potentially bite us when threading, but then again threads and concurrency and shared resources should always already put you on high alert. It also solves the inefficiency issue, and does not require you to make defensive copies or anything else that requires reading specific instructions!

XapaJIaMnu:
I was thinking about this in the past few days. How many repos is a user going to have? 2-3? 5? 10? The map size would be tiny and the memory copy overhead would be minimal.

Option 3 makes it inherently not thread safe unless we get locks involved. Getting the map referece/pointer could come together with a read lock that gets discarded when it goes out of scope (getting the map would be a more complicated as). Updating the setting would involve acquiring that lock. You are right at the moment we don't have any codepaths that have the potential of hitting this race condition so this might not be necessary.

On the other hand option 1 requires you to make defensive copies for map lookups which could also be error prone, as shown.

Another option is to have a specialised method for the map setting type that performs the lookup and forbids the user from using it directly, but that makes the setting class more confusing to use.

I am leaning a bit more towards 1) but I acknowledge that the mechanism is a bit error prone and that in the short term this might cause more issues than a potential race condition in 3). I don't think there are any performance consideration worth noting in this case.

jelmervdl:
The issues caused by (1) should be more easily detectable than the ones caused by (3). Honestly I was a bit surprised that clangd didn't flag it. Maybe it would have if it had been a `std::unorderd_map` instead of a `QMap`…

But anyway. Lets then stick to (1) for now! Is that this done then? (Except for the apparent merge conflict?)

jelmervdl:
Oh sorry I missed this bit:

> I was thinking about this in the past few days. How many repos is a user going to have? 2-3? 5? 10? The map size would be tiny and the memory copy overhead would be minimal.

Yep, the size of the copies would be small. But the number of copies created would scale with the number of models in the Languages tab or in the dropdown.

But in the short term, even in the worst case that would be only a couple of 100? I don't think you'll notice. If it does become a performance issue we can tackle it then?

XapaJIaMnu:
> But anyway. Lets then stick to (1) for now! Is that this done then? (Except for the apparent merge conflict?)
I think that resolved the conflict. Let's build and retest again. If it works i'll merge it. Next step: Single column language selection -> 2 Column language selections.

>  But in the short term, even in the worst case that would be only a couple of 100? I don't think you'll notice. If it does become a performance issue we can tackle it then?
Yeah it could be a couple of hundred copies but it is such a small amount of data that it'd be kept in CPU cache. True, we can fix it once it becomes a performance issue.

XapaJIaMnu:
`QVariant::save: unable to save type 'QSettingsMap' (type id: 1067).`. I think the automatic merge failed.

XapaJIaMnu:
@jelmervdl at the moment adding a new repo doesn't trigger "FetchRemoteModels". Is this a bug, or did we change the default behaviour?

jelmervdl:
It is intentional. See the comment here: https://github.com/XapaJIaMnu/translateLocally/pull/102#discussion_r910448924

PR #113, opened 2022-07-20
Handle local and remote models from the languages tab

XapaJIaMnu:
As we will have more and more languages, it will not be easy to select them for downloading from the dropdown menu. Instead, use a separate window for it. I'm reusing the languages tab with some colour coded text:
Screenshot_20220720_221353
Screenshot_20220720_221439

This is WIP with a few issues needing to be ironed out before  it makes it into master.

1. ~Newly downloaded models do not correctly remove the model from the "Available for download" model list. This might have been broken at some point. Restarting translatelocally fixes it. **This is not a new issue, it's broken in master as well.**~
2. ~Upgrading a model doesn't always correctly remove outdated model from the config directory (**also occurs in master**).~
3. ~Upgrading a model doesn't correctly mark the newly downloaded model as upgraded (likely due to 2) and the user has the option to upgrade it again.~
4. Users should be able to sort the models to be downloaded by column so that searching is easier.

Taking suggestions for UI and such.


kpu:
I think `-1` means not present.  The columns for installed and version could just be combined.  

The primary use case is pick a language pair (which by the way should be the two dropdowns) and download what's needed.  Secondary is pick a type (which in theory is a ranked list of priorities, though many will just go with the default; people with custom models may want different stuff).  Another big use case is check for new models and update everything.  I think the UI has to be oriented to that.  Advanced model-level updating can be there.

XapaJIaMnu:
So, we had a grand plan with @jelmervdl to work towards having two separate drop downs that select language on the main window. That way we can also make pivoting work quite easy, but unfortunately it makes downloading models really messy. (At least the way we imagined it beforehand).

So as a prerequisite, we wanted to create a separate window where the user can download models and the main application window is reserved just for translations and language selection.

You are right though, that the most straightforward way for the user is to select two languages and then let translateLocally figure out what needs to be downloaded.

kpu:
If we wanted to, we could put a download icon next to a language that needs a download in the selection.

jelmervdl:
> If we wanted to, we could put a download icon next to a language that needs a download in the main screen selection.

I like the idea of making downloading close to the default behaviour (but give it a prompt before actually starting the one-time download). For this to work, we do need to download the model lists though, just like now. That could still be a "Download more…" option at the bottom of both dropdown menus.

> So, we had a grand plan with @jelmervdl to work towards having two separate drop downs that select language on the main window. 

I think @XapaJIaMnu has the notes we took somewhere. But yes, the idea was two dropdowns for languages on both sides. I don't remember what we thought about the idea of having the destination language dropdown have submenus if there were multiple models (e.g. base, tiny11) or pivoting "routes" to come to that src-trg language pair, or whether that was too messy.

Esp. the multiple pivoting routes could become very large easily. Maybe for that, I'd rather have an optional third dropdown in the middle for pivoting language that can be toggled, and then have that do the submenus for model type as well. Making pivoting explicitly opt-in also makes it a bit more explicit that you're opting into lower quality. It can be toggled in the settings with some description about how and why, and maybe also one of the menus. And we can add a special option for "no pivoting"/"direct" so you can enable it visually but not use it all the time.

> So as a prerequisite, we wanted to create a separate window where the user can download models and the main application window is reserved just for translations and language selection.

The separate window is still a nice thing to have for advanced users searching for a specific model I think. It is searchable and sortable. It also gives us a lot more space to show what the difference is between the models. Much more than a menu item.

src/inventory/ModelManager.cpp
```diff
@@ -741,46 +759,114 @@ QVariant ModelManager::headerData(int section, Qt::Orientation orientation, int
...
+                } else {
+                    return QColor(0xE7, 0xE8, 0xE6);
+                }
+            case Qt::ForegroundRole:
+                if (model.outdated()) {
+                    return QColor(0xFD, 0x25, 0x58);
+                }
+            default:
+                return QVariant();
+        }
```
jelmervdl, line 847: The indent is different here from the other switch cases (Sorry I hate leaving these comments :( )

src/inventory/ModelManager.cpp
```diff
@@ -741,46 +759,114 @@ QVariant ModelManager::headerData(int section, Qt::Orientation orientation, int
...
+    Model model; // Make sure we have all local models before the remote ones
+    if (index.row() < localModels_.size())
+        model = localModels_[index.row()];
+    else if (index.row() < localModels_.size() + newModels_.size())
+        model = newModels_[index.row() - localModels_.size()];
 
     if (role == Qt::UserRole)
         return QVariant::fromValue(model);
 
     switch (index.column()) {
```
jelmervdl, line 794: There seems to be quite a bit of duplicate code here because we switch on column first, and then on role. Maybe flip that around?

src/settings/TranslatorSettingsDialog.cpp
```diff
@@ -198,3 +215,46 @@ void TranslatorSettingsDialog::on_deleteRepo_clicked()
...
+        // Can only download one model at a time for now
+        QMessageBox::warning(this, tr("Warning"), "Can only download one model at a time for now. Please select just one model.");
+    } else {
+        Model oldmodel = modelManager_->data(ui_->localModelTable->selectionModel()->selectedIndexes().first(), Qt::UserRole).value<Model>();
+        // Find the model in the updated models list
+        auto todownload = modelManager_->findModelForUpdate(oldmodel);
+
+        if (todownload) {
+            emit downloadModel(todownload.value());
+            this->setVisible(false); // Hide the settings window so we can see the download.
```
jelmervdl, line 251: I suspect it makes more sense to show download progress either in a separate window (like Finder or any file manager when copying files) or in the settings window itself. But for now, while only allowing downloading *one* model, this is still acceptable behaviour? Especially because then it also selects that model and is ready to be used, which is nice.

src/mainwindow.cpp
```diff
@@ -306,8 +308,11 @@ void MainWindow::handleDownload(QFile *file, QString filename, QVariant extra) {
     ModelMeta meta = extra.value<ModelMeta>();
     meta.installedOn = QDateTime::currentDateTimeUtc();
     auto model = models_.writeModel(file, meta, filename);
-    if (model) // if writeModel didn't fail
+    if (model) { // if writeModel didn't fail
         settings_.translationModel.setValue(model->path, Setting::AlwaysEmit);
+        models_.updateAvailableModels(); // @TODO This doesn't remove the newly downloaded model from the the list of
+                                         // of available models for downloading at this time. Need to fix it.
+    }
```
jelmervdl, line 315: Weird… `writeModel()` already calls `updateAvailableModels()` at the end.

XapaJIaMnu: Yeah i noticed that, but even in master it doesn't work. Not exactly sure what goes wrong in there.

src/settings/TranslatorSettingsDialog.cpp
XapaJIaMnu: We can make batch downloading at a later point I think. My goal is to make this window a one-stop model management, so we can then refactor the main window without caring as much about how to display models for download (even if we eventually need to account for that)

src/mainwindow.cpp
jelmervdl: I've made #114 for it because it looks like a logic bug somewhere (I suspect the downloaded local model is somehow slightly different from the scanned-at-startup local model and so the remote-local match-up fails?)

src/settings/TranslatorSettingsDialog.cpp
jelmervdl: That sounds good to me.

src/inventory/ModelManager.cpp
XapaJIaMnu: Fixd

XapaJIaMnu:
Do you think we should also implement searchable/sortable model list before marking this as ready to review?

jelmervdl:
Going by that screenshot it does look like a useful addition. I already have sorting working, now trying to add a filter field (and am mostly there)

jelmervdl:
image

It is not yet the prettiest but it is functional. Cosmetics can wait for a different pull request. Still contemplating whether I should have put it outside the scroll area and made the table smaller when it pops up either above or below it, or whether to keep the hovering implementation. But this was the simplest one to implement.

jelmervdl:
Other than that the only thing I think we can improve is replacing the "Installed" column with just hiding the `-1` version number. Then that column plus the row backgrounds tell you whether it is local or not. Sorting on that column still also moves local ones to top/bottom.

That change will save us a column, which makes the need for horizontal scrolling more of an exceptional case.

src/inventory/ModelManager.h
```diff
@@ -107,6 +113,14 @@ struct Model : ModelMeta {
...
     }
 
+    inline QString getReportedRepo() const {
+        // Models could come with their own repository specified in model_info.json
+        // However it is more reliable to remember where one model came from, by setting the
+        // repository URL when downloading. That being sad, sometimes it is useful to specify this
+        // for builtin models so that we remember that this is a self managed model
+        return repo;
+    }
+
```
jelmervdl, line 123: Why does this one get a getter? Because you want to be able to specify a default value for local models that don't have the info in model_info.json?

jelmervdl: Aah, okay, makes sense!

XapaJIaMnu:
I like it. Several minor things:
1) By default I think we should be sorting by installed locally.
2) I don't like the three fields local Version/ Remote Version/Installed. It takes too much space and it's clunky. Is there a better way to do it?
3) Should there be a resize-width-to-fit for the table?
4) Add a button for the prompt? `ctrl+f` is intuitive but maybe someone would miss it.

I'd like to freeze on the feature addition here as this is already a really fat PR.

XapaJIaMnu:
Seems like I missed your comment.
How about adding an "advanced" button that adds the precise version numbers and maybe a few other columns.
Then local/remote/installed can be just installed which can be true/false/outdated?

XapaJIaMnu:
1) Low priority, since by default you'd only see the extra models once you fetch external models.
2) I agree about extra models... Do you want to collapse download/update into one. It can change text based on the model selected.
3) That can be done in the front end once we have two column drop downs and it wouldn't be so bad.
4-5) I like "View" and select columns idea.

jelmervdl:
I'm leaving these for another pull request, I think:
1. Sort downloaded models on top
2. Make download button change to "update" if they're updating  
(But what if you select two models, one to download, one to update? And is the distinction here that important?)
3. Menu items for toggling column visibility
4. Menu item for find in table.  
(Both menu items still have the issue that there are multiple tables in this window, but luckily only one is visible at a time so we know which table the menu items should act on.)

Other than that, I think this is pretty much ready!

XapaJIaMnu:
I agree that those are low priority and I will open an issue to track them. Will review the changes and merge later tonight.

src/settings/TranslatorSettingsDialog.ui
```diff
@@ -17,7 +17,7 @@
    <item>
     <widget class="QTabWidget" name="tabWidget">
      <property name="currentIndex">
-      <number>0</number>
+      <number>1</number>
```
XapaJIaMnu, line 20: What's happening here? Is that accidental or?

XapaJIaMnu: Ok, appears I have changed it... No idea what it does, as the change was made through the GUI.....

jelmervdl: I think it is the tab that is selected. So this would entail the default tab would be the Languages tab? I've reverted it.

jelmervdl:
Eh, discovered one more issue tonight as my macbook switched to dark mode in the evening, hehe:
image

I'll have a look at blending colours instead of overriding them. QColor should have some facilities for that I'd expect.

XapaJIaMnu:
Oh god, I forgot the bloody dark mode...

We could have some dark mode scheme we could make use of, potentially?

jelmervdl:
Screenshot 2022-07-25 at 22 56 19
Screenshot 2022-07-25 at 22 56 30

Not 100% aesthetically pleasing I must say, but at least the model text is readable again. Should be pretty portable since it blends with the background colour from the system palette.

XapaJIaMnu:
It's not idea, but I think it is a net improvement. The colour blending code is quite small and I don't see it as too harmful to be in the `ModelManager` for now.

PR #115, opened 2022-09-12
Compile for generic x86-64 which is compatible with M1 and M2

jelmervdl:
Basically taken from https://stackoverflow.com/questions/69803659/what-is-the-proper-way-to-build-for-macos-x86-64-using-cmake-on-apple-m1-arm but I tested a local build with these options and it worked on an M2 Macbook Air.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -35,6 +35,9 @@ jobs:
           - name: "macOS Big Sur 11.0 avx"
             os: macos-11.0
             arch: "core-avx-i"
+          - name: "macOS Big Sur 11.0 x86-64"
+            os: macos-11.0
+            arch: "x86-64"
          # avx2 build is not any faster, since the critical codepath (intgemm) already uses the most optimal codepath for the user's PC
          # - name: "macOS Catalina 10.15 avx2"
          #   os: macos-10.15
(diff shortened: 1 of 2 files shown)


XapaJIaMnu:
And it translated? We don't have generic intgemm path for it?

XapaJIaMnu:
Should we add a generic x86 build for windows and ubuntu and label them compatibility?

jelmervdl:
I'm not entirely sure which Intgemm path it takes… but yes it works properly:
189628664-5a90b585-2919-4927-8622-191bc04f8b51

.github/workflows/build.yml
```diff
@@ -117,7 +120,7 @@ jobs:
       # and build directories, but this is only available with CMake 3.13 and higher.  
       # The CMake binaries on the Github Actions machines are (as of this writing) 3.12
       run: |-
-        cmake $GITHUB_WORKSPACE -DCMAKE_BUILD_TYPE=$BUILD_TYPE -DBUILD_ARCH=${{ matrix.arch }} ${{ env.ccache_cmake }}
+        cmake $GITHUB_WORKSPACE -DCMAKE_BUILD_TYPE=$BUILD_TYPE -DBUILD_ARCH=${{ matrix.arch }} -DCMAKE_OSX_ARCHITECTURES=x86_64 ${{ env.ccache_cmake }}
```
XapaJIaMnu, line 123: Is this necessary on top of build arch?

PR #122, opened 2022-10-13
try to use qt6 for windows

XapaJIaMnu:
Qt6 didn't use to work for windows, let's try again

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -350,8 +350,8 @@ jobs:
       working-directory: C:\vcpkg
       run: |
         $Env:VCPKG_BUILD_TYPE = 'release'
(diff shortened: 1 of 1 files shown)


XapaJIaMnu:
Someone needs to Fix VCPKG Qt6 linguist build likely in a similar manner as the qt5 fix. Windows machine necessary cause VCPKG development from githubCI is not fun...
https://github.com/microsoft/vcpkg/issues/2834
https://github.com/microsoft/vcpkg/issues/20244

XapaJIaMnu:
Update on this, the qt6 build would succeed if we install `qtsvg:x64-windows`. Unfortunately, it picks up a dynamic build, and even with all dlls placed correctly, the GUI just looks shit (platform plugin not working well). 
Also, it seems to not pick up SSL when doing the qt6 build, but the qt5 picks it up fine.
For now keep using qt5 on windows.

Furthermore, it doesn't need protobuf installed via vcpkg

PR #125, opened 2022-11-30
Fallback ListModels when offline

### LINKED ISSUE #124: ListModels request which includes includeRemote=true should not fail when offline
When looking into https://github.com/jelmervdl/translatelocally-web-ext/issues/62#issuecomment-1332349187 I noticed that TranslateLocally doesn't gracefully handle the offline scenario. That should be fixed.

jelmervdl:
When a ListModels request is handled with `includeRemote=True` but we're offline (or fetching fails for some reason) gracefully just return the local list of models.

I considered passing the error onto the caller, but the ListModels request can still partially succeed and serve its purpose. So falling back gracefully was the route I chose.

Fixes #124.

### DIFF
--- a/src/cli/NativeMsgIface.cpp
+++ b/src/cli/NativeMsgIface.cpp
@@ -204,13 +204,22 @@ void NativeMsgIface::handleRequest(TranslationRequest request) {
 void NativeMsgIface::handleRequest(ListRequest request)  {
     // Fetch remote models if necessary.
     if (request.includeRemote && models_.getRemoteModels().isEmpty()) {
+        std::shared_ptr<ListRequest> requestCopy(std::make_shared<ListRequest>(request));
+
+        // If the fetch fails, make sure remote is excluded next time.
+        QMetaObject::Connection errorConnection = connectSingleShot(&models_, &ModelManager::error, this, [this, requestCopy](QString error) {
+            requestCopy->includeRemote = false;
+        });
+
         // Note: this might pick up the completion of an earlier fetchRemoteModels()
         // request but that's okay since fetchRemoteModels() returns early if a
         // fetch is still in progress. Also, fetchedRemoteModels() is called
         // regardless of whether errors occurred during the fetching.
-        connectSingleShot(&models_, &ModelManager::fetchedRemoteModels, this, [this, request]([[maybe_unused]] QVariant ignored) {
-            handleRequest(request);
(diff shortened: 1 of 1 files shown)


src/cli/NativeMsgIface.cpp
```diff
@@ -204,13 +204,22 @@ void NativeMsgIface::handleRequest(TranslationRequest request) {
...
+        });
+
         // Note: this might pick up the completion of an earlier fetchRemoteModels()
         // request but that's okay since fetchRemoteModels() returns early if a
         // fetch is still in progress. Also, fetchedRemoteModels() is called
         // regardless of whether errors occurred during the fetching.
-        connectSingleShot(&models_, &ModelManager::fetchedRemoteModels, this, [this, request]([[maybe_unused]] QVariant ignored) {
-            handleRequest(request);
+        connectSingleShot(&models_, &ModelManager::fetchedRemoteModels, this, [this, requestCopy, errorConnection]([[maybe_unused]] QVariant ignored) {
+            QObject::disconnect(errorConnection);
```
XapaJIaMnu, line 219: Why do we need a disconnect here? Isn't errorConnection single shot?

jelmervdl: Yep, but if fetchedRemoteModels triggers without any errors, it will still be there once fetchedRemoteModels finished.

Maybe it would have been better if fetchedRemoteModels triggers in all cases, but when an error occurs it triggers with different arguments. That would remove this problem of catching either signal.

src/cli/NativeMsgIface.cpp
```diff
@@ -204,13 +204,22 @@ void NativeMsgIface::handleRequest(TranslationRequest request) {
 void NativeMsgIface::handleRequest(ListRequest request)  {
     // Fetch remote models if necessary.
     if (request.includeRemote && models_.getRemoteModels().isEmpty()) {
+        std::shared_ptr<ListRequest> requestCopy(std::make_shared<ListRequest>(request));
```
XapaJIaMnu, line 207: Why are we not working on the requests directly?

jelmervdl: The function will have finished and `request` has gone out of scope by the time one of the callbacks is called. I copied it over to a shared_ptr so it can share a single request between the two signal handlers. Previously we just copied (or moved) the request into the one signal handler.

PR #128, opened 2023-01-18
ARM build

XapaJIaMnu:
Maybe a small todo, to have an arm build CI. I'll add one soon

Since there's no way to make CMAKe do something after the generator step (at least not cleanly), currently we have:

```
Added subdirectory /home/ubuntu/translateLocally-arm-new/3rd_party/bergamot-translator/3rd_party/ssplit-cpp/src
CMake Warning at CMakeLists.txt:333 (message):
  Building on ARM.  You need to manually fix the broken build due to RUY
  using bad header names and QT's MOC.  Please run:


/home/ubuntu/translateLocally-arm-new/cmake/fix_ruy_build.sh /home/ubuntu/translateLocally-arm-new /home/ubuntu/translateLocally-arm-new/build
CMake Warning at CMakeLists.txt:335 (message):
  Before building the project
```

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -301,6 +301,64 @@ jobs:
       with:
         name: translateLocally.latest.${{ matrix.os }}.${{ matrix.arch }}.deb
         path: ${{github.workspace}}/build/deb/*.deb
+        
+  build-arm:
+    strategy:
+        matrix:
+          include:
+            - name: "Ubuntu 22.04 ARM"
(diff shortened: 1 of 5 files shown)


XapaJIaMnu:
The Windows 2022 build is broken, due to something that happened on windows side. Not sure how to debug it, but everything else works, plus I added extra runners, and an arm test on a self hosted runner! The only thing that's left is a dedicated M1 build.

PR #131, opened 2023-05-04
Improve dependency management

XapaJIaMnu:
Pr where we update workflows


XapaJIaMnu:
Now that i finally fixed windows compilation I'll merge this. Will fix mac next week.

PR #141, opened 2023-08-24
inventory: include shared models directories

### LINKED ISSUE #139: Model directory path
Currently, models are downloaded into the dir `~/.config/tranlateLocally/`.

1. This is not ideal for people tracking their dotfiles with .git for example.
   A better place for downloadable binary data would be under `$XDG_DATA_HOME`.

2. It would be nice to also look for models in each `$XDG_DATA_DIRS`.
   This would allow system package managers to install models for all users
   (easier deployment and updates).

pacien:
This scans for shared models installed through the system package
manager, located in all of $XDG_DATA_DIRS/share/bergamot/models.

Github: closes #139

### DIFF
--- a/src/inventory/ModelManager.cpp
+++ b/src/inventory/ModelManager.cpp
@@ -7,6 +7,7 @@
 #include <QDirIterator>
 #include <QFile>
 #include <QSaveFile>
+#include <QStandardPaths>
 #include <QFileInfo>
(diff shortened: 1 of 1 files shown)


pacien:
Works nicely with: https://github.com/NixOS/nixpkgs/pull/251119

XapaJIaMnu (approved):

PR #142, opened 2023-08-24
inventory: use standard AppDataLocation for model files

### LINKED ISSUE #139: Model directory path

pacien:
inventory: use standard AppDataLocation for model files

This puts the models in $XDG_DATA_DIR/translateLocally on Linux,
or equivalent user data directories on other platforms.

This avoids putting relatively large files in the config directory,
which may be versioned for user dotfiles.

Github: closes #139

### DIFF
--- a/src/inventory/ModelManager.cpp
+++ b/src/inventory/ModelManager.cpp
@@ -94,7 +94,8 @@ ModelManager::ModelManager(QObject *parent, Settings * settings)
     QDir configDir = QStandardPaths::writableLocation(QStandardPaths::AppConfigLocation);
     QDirIterator configDirIter = QDirIterator(configDir);
     while (configDirIter.hasNext()) {
-        QFileInfo fileInfo = configDirIter.nextFileInfo();
+        configDirIter.next();
+        QFileInfo fileInfo = configDirIter.fileInfo();
         if (fileInfo.isDir() || fileInfo.fileName().endsWith(".tar.gz")) {
             QFile::rename(fileInfo.absoluteFilePath(), appDataDir_.filePath(fileInfo.fileName()));
         }
--- a/src/inventory/ModelManager.cpp
+++ b/src/inventory/ModelManager.cpp
@@ -78,9 +78,7 @@ ModelManager::ModelManager(QObject *parent, Settings * settings)
     , settings_(settings)
     , isFetchingRemoteModels_(false)
 {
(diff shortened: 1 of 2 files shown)


XapaJIaMnu (approved):
I don't think we should try to move anything, that could potentially get tricky. Scanning one additional directory for models on startup is not a big deal.

pacien:
Quoting Jelmer (2023-08-25 09:58:47)
> One thing that's unclear to me is what happens with the models people
> have already downloaded? 
> 
> Will they still show up? (The scanning code seems to purely be an
> addition, so yes?)

In the current state of this pull request, they won't show up anymore.
But I can add the previously used configuration directory as scanned path.

> Or should they be moved on first run? (Sounds tricky to make fault tolerant)

This should be fine since the target directory is a fresh one for now.
Except if I'm missing something?

XapaJIaMnu:
Yes, please add  scanning the config dir as well. I think it was here.
https://github.com/XapaJIaMnu/translateLocally/blob/0c0bc1858623f44553791606f4382d12a9c00c74/src/inventory/ModelManager.cpp#L503

pacien:
Quoting Nikolay Bogoychev (2023-08-25 13:41:26)
> Yes, please add  scanning the config dir as well. I think it was here.
> https://github.com/XapaJIaMnu/translateLocally/blob/0c0bc1858623f44553791606f4382d12a9c00c74/src/inventory/ModelManager.cpp#L503

Done, and also rebased to resolve conflict.

pacien:
I added a commit which moves the model directories and tarballs from the config
directory to the new app data directory. This should work fine in most cases.
If the move operation fails (silently), the models are still listed because
the config directory is still scanned.

I'll let you decide whteher to pick this latest commit or not.

XapaJIaMnu:
Hmm, the windows build is suddenly broken. I need to investigate to see if something changed in the github runner  or inadvertently this broke something (Last week it was passing).

The Mac runner had some brew changes that happened that I need to fix as well...

Issue #140, opened 2023-08-23
Crash with SIGABRT on translation attempt


pacien:
Context: I'm trying to create a distro package of this program for NixOS.
(https://github.com/NixOS/nixpkgs/pull/251051).

The CLI and GUI start and are able to donwload models.

However, attempting a translation results in a crash with SIGABRT.
This is happening both with and without MKL, and both on CLI and in GUI.

Platform info: NixOS 23.05 (Linux 6.1 LTS, x86_64).
translateLocally commit: 9f3195.

```fish
❯ nix build github:NixOS/nixpkgs/f906ee5bf2816bd9817a86355051339d84bdfaff#translate-locally
# built output is placed in ./result

❯ ./result/bin/translateLocally -d en-fr-tiny
Downloading English-French type tiny...
100% [############################################################]
Model downloaded successfully! You can now invoke it with -m en-fr-tiny

❯ echo "hello world" | ./result/bin/translateLocally -m en-fr-tiny
fish: Process 516219, './result/bin/translateLocally' from job 1, 'echo "hello world" | ./result/b…' terminated by signal SIGABRT (Abort)
```

Using the `--debug` flag does not give any additional info.

XapaJIaMnu:
Did you have mkl installed before compiling? Most likely it was compiled without compatible Blas library. We should make it hard fail in this case.

pacien:
Quoting Nikolay Bogoychev (2023-08-24 00:33:50)
> Did you have mkl installed before compiling? Most likely it was compiled
> without compatible Blas library. We should make it hard fail in this case.

I'm producing two versions of the package: one with mkl and one without.
Both are compiling fine, but both are crashing when translating.
I also tried to add Blas (version 3) as dependency for both, but same crash.

XapaJIaMnu:
During compilation do you have something like "found mkl" in the cmake log?

pacien:
Quoting Nikolay Bogoychev (2023-08-24 01:02:16)
> During compilation do you have something like "found mkl" in the cmake log?

CMake seems to be able to find all the other dependencies automatically,
but not BLAS, CBLAS, and MKL, even though they are available at the standard
include paths.

I see that there are some custom FindCBLAS.cmake and FindMKL.cmake in the
submodules specifically for that. Is there a way to bypass those?

pacien:
This seems to be enough to fix the BLAS and CBLAS discovery:

  cmakeFlags = [
[...]

PR #144, opened 2023-08-25
readme: more explicit list of build dependencies

pacien:
It wasn't clear that it could be used without Intel MKL,
using BLAS instead.

Github: relates to #140

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,7 +16,20 @@ make -j5
 ./translateLocally
 ```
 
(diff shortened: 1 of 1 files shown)


XapaJIaMnu (changes requested):
We really do not want to encourage the use OpenBLAS in place of MKL. OpenBLAS is multiple times slower regardless of the hardware used (AMD or Intel). Also, depending on how OpenBLAS is compiled, it will try to auto parallelise GEMM operations with `omp` but our matrices are way too small for this to actually be helpful resulting in extreme cases in 100x slowdowns (Think  multicore processors). (OMP can be controlled through the `OMP_NUM_THREADS` environment variable but we do not expect the user to be aware to set that.

I will reword the README to reflect that. We should also allow linking to shared mkl libraries on Linux.

PR #168, opened 2024-05-12
M1 Build attempt

XapaJIaMnu:
First attempt to get M1 build on the workflows

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -128,7 +128,7 @@ jobs:
         # and build directories, but this is only available with CMake 3.13 and higher.
         # The CMake binaries on the Github Actions machines are (as of this writing) 3.12
         run: |-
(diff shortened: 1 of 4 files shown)


muzzah:
Just wanted to confirm, does the ARM Mac releases still require rosetta at the moment? I prefer not to download rosetta and the Mac release is asking me to download it still

XapaJIaMnu:
We still don't have an automated m1 build. Github offers m1 runners but I don't have access to m1 hardware myself to debug it and set it up.

For now native m1 build is only available if you build it yourself, sorry;/

jelmervdl:
@XapaJIaMnu I have a personal M3 MacBook now, when I find some time I can give it a go.

jelmervdl:
Build seems to be working, signing fails though. Definitely an expired key, but also the arm build has some parse errors of which I don't know the source.

